PR #3, opened 2017-08-02
Check if a file exists before trying to load it.

chrismullins:
If the user attempts to load a file which is not there, show a critical
QMessageBox with a short message and the full expected filepath.  For example,
this code will run if  a user tries to load a file from the
'Open recent' menu which has been renamed or moved.

If the file doesn't exist, return immediately.

TODO: Log an error here.

### DIFF
--- a/mnelab.py
+++ b/mnelab.py
@@ -147,11 +147,9 @@ class MainWindow(QMainWindow):
             File name.
         """
         if not exists(fname):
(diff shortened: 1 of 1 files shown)


cbrnr:
Thanks @chrismullins, I've added a commit to remove the file from the recent menu

chrismullins:
Good call!

PR #4, opened 2017-08-04
WIP: Improve the dialog for picking bad channels.

chrismullins:
Currently, the channel picker dialog does a good job of allowing the user to
choose bad channels, but has a few shortcomings:

(1) It's difficult to see, at a glance, which channels have already been marked
as bad.

(2) When the user begins clicking on new channels to mark as bad, any
previously marked channels become unselected. Thus, making sure you've selected
everything you want requires careful scrolling.

Using two list widgets to shuffle channels goes some lengths to solve both
these issues.


cbrnr:
A new channel properties dialog is available in master.

PR #6, opened 2017-08-08
Fix montage crash

chrismullins:
Fix two crashes in the montage workflow + a few flake8 fixes.

This fix introduces the issue that, between differing versions of mne, the `Montage.plot()` function creates entirely different plots.  I'll see if I can come up with an elegant way to combine the setting and the viewing of the montages.  This might be a useful feature for someone who didn't know which montage they needed to use, but could figure it out through looking at the configuration and lining up the channel names. (esp. those who are entirely new to EEG)  

Also TODO: allow the user to load their own montage into the selection menu from a file.

### DIFF
--- a/mnelab.py
+++ b/mnelab.py
@@ -388,7 +388,7 @@ class MainWindow(QMainWindow):
             self.all.current.montage = name
             self.all.data[self.all.index].montage = name
             montage = mne.channels.read_montage(name)
-            try: 
+            try:
                 self.all.current.raw.set_montage(montage)
                 self.all.data[self.all.index].raw.set_montage(montage)
                 self._update_infowidget()
--- a/mnelab.py
+++ b/mnelab.py
@@ -388,10 +388,13 @@ class MainWindow(QMainWindow):
             self.all.current.montage = name
             self.all.data[self.all.index].montage = name
             montage = mne.channels.read_montage(name)
-            self.all.current.raw.set_montage(montage)
-            self.all.data[self.all.index].raw.set_montage(montage)
-            self._update_infowidget()
-            self._toggle_actions()
+            try: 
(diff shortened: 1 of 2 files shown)


mnelab.py
```diff
@@ -428,7 +431,10 @@ def plot_psd(self):
 
     def plot_montage(self):
         montage = mne.channels.read_montage(self.all.current.montage)
-        montage.plot(kind="topomap")
+        if int(mne.__version__.split('.')[1]) >= 15:
```
cbrnr: I'd prefer a a try/except block that catches the TypeError when you specify the new keyword argument.

mnelab.py
```diff
@@ -388,10 +388,13 @@ def set_montage(self):
             self.all.current.montage = name
             self.all.data[self.all.index].montage = name
             montage = mne.channels.read_montage(name)
-            self.all.current.raw.set_montage(montage)
-            self.all.data[self.all.index].raw.set_montage(montage)
-            self._update_infowidget()
-            self._toggle_actions()
+            try:
```
cbrnr: Here a check if there is at least one channel name in the data and montage would probably be better than a try/except block. Then a dialog should pop up with an error message.

cbrnr:
Fixed in master. I've added a dialog with an error message if no name matches. Also, I'm not using the `kind` argument anymore, which means that it works for all MNE versions (but for MNE < 0.15 a 3D plot will be shown, whereas for MNE >= 0.15 a topomap plot will be shown - topomap montage plots were introduced with 0.15).

cbrnr:
I've also just pushed the ability to view montages from the dialog to master.

chrismullins:
Sounds good!

PR #9, opened 2017-08-26
WIP: Add the Edit Channel Info dialog box with selection logic.

chrismullins:
NOT READY FOR MERGE

TODO: If the dialog's end result is accepted, this should write the
results back to the channel object and record what was done.

TODO: Add an "undo" button and keep track of changes within one
session with the dialog open.


cbrnr:
A new channel properties dialog is available in master.

PR #8, opened 2017-08-22
WIP: Edit Channels dialog, allowing bulk editing of channel data

chrismullins:
NOT READY FOR MERGE.  Only for discussion purposes.

This doesn't manipulate any data yet, just demonstrates the proposed mechanics of the UI.
Allows editing of channel types and bad channels individually, or altogether through
a separate Edit Selected Channels dialog.

Discussion should take place in the #5 Issue.

PR #12, opened 2017-12-20
[WIP] Channel properties dialog

### LINKED ISSUE #5: Improve the Bad Channels picker dialog box
From the associated PR #4:

> Currently, the channel picker dialog does a good job of allowing the user to
> choose bad channels, but has a few shortcomings:
> 
> (1) It's difficult to see, at a glance, which channels have already been marked
> as bad.
> 
> (2) When the user begins clicking on new channels to mark as bad, any
> previously marked channels become unselected. Thus, making sure you've selected
> everything you want requires careful scrolling.

This change introduces a few new files (including a `.ui` file from QtDesigner and generated `.py` from the `pyuic5` tool) and @cbrnr might have preferences about how this sort of thing is organized/implemented.  Since this might entail multiple PR's, an issue might be a better place for discussion.

Please feel free to review my code and/or let me know your thoughts and ideas.

cbrnr:
Fixes #5, #4, #8, #9.

screen shot 2017-12-21 at 16 28 27

Done:
- [x] Columns are are nicely distributed (auto-size)
- [x] All columns can be sorted by clicking on the header
- [x] Bad channels can be toggled
- [x] Removed "Set bad channels" dialog
- [x] Labels can be edited
- [x] Update data in Bad column based on checked/not checked
- [x] The Type column contains comboboxes with all supported channel types (cf https://github.com/mne-tools/mne-python/pull/4842)
- [x] Types can be changed

Doesn't work yet:
- [ ] Only the popup should be visible when editing a type, the combobox widget itself should not be drawn
- [ ] The checkboxes in the Bad column should probably be centered (minor)

Up for discussion:
- [ ] I've disabled multi-row selection because I haven't found an intuitive way to implement this. Actually, I think this is not necessary since selecting individual bad channels is quick and intuitive. Both shortcomings mentioned in #5 are addressed (IMO). However, suggestions are welcome.

@chrismullins sorry for neglecting your PRs up until now, I didn't have the time to work on MNELAB.


cbrnr:
Here's a good link for the discussion point - we'd definitely need a new column with checkboxes for selecting rows: https://ux.stackexchange.com/questions/42043/how-to-avoid-repetitive-actions-in-table-rows

PR #15, opened 2018-01-23
Compute ICA

cbrnr:
I seem to have hit a dead end here. Here's what I want to do: To run ICA, you can click on "Tools" - "Run ICA..." and select some options in the dialog (make sure that you have loaded a dataset before). After confirming with OK, a new dialog pops up, which provides an option to cancel the ongoing computation.

I cannot start `ica.fit()` in the main Python process, because it would block the GUI (which I only need to be able to cancel the calculation). I asked a [question on SO](https://stackoverflow.com/questions/48262966/terminate-a-long-running-python-command-within-a-pyqt-application), and they recommended the multiprocessing module. However, this approach has two drawbacks:

1. It only works for the toy example and not for the real thing in MNELAB (on macOS, I get dozens of these messages: "The process has forked and you cannot use this CoreFoundation functionality safely. You MUST exec(). Break on _THE_PROCESS_HAS_FORKED_AND_YOU_CANNOT_USE_THIS_COREFOUNDATION_FUNCTIONALITY___YOU_MUST_EXEC__() to debug." On Linux, I get a similar error: "Gdk-Message: python: Fatal IO error 11 (Resource temporarily unavailable) on X server :1." It seems like the other process (the ICA calculation) is trying to change the GUI, which I don't understand at all.
2. Even if this worked, it consumes far too much memory. I tried this with a moderately-sized EEG file that consumes about 800MB in memory. The other process grows up to 3GB in size. This is unacceptable.

Threads would be much better suited for this use case, because only one Python process is running at a time anyway. I just want the GUI (more precisely the dialog window with the Cancel button) to stay responsive so that the computation can be interrupted. However, it seems like threads cannot be killed from outside. Instead, a sentry variable should be modified and the running thread should periodically check for this variable in order to decide whether to continue or quit. I can't do this since I don't want to change anything in the ICA calculation code.

Help is very much appreciated.

### DIFF
--- a/mnelab.py
+++ b/mnelab.py
@@ -2,14 +2,17 @@ import sys
 import matplotlib
 from PyQt5.QtWidgets import QApplication
 from mnelab import MainWindow
+import multiprocessing as mp
 
 
-matplotlib.use("Qt5Agg")
-app = QApplication(sys.argv)
-app.setApplicationName("MNELAB")
-app.setOrganizationName("cbrnr")
-main = MainWindow()
-if len(sys.argv) > 1:  # open files from command line arguments
-    for f in sys.argv[1:]:
-        main.load_file(f)
-sys.exit(app.exec_())
+if __name__ == "__main__":
+    mp.set_start_method("spawn")  # required for Linux/macOS with PyQt
+    matplotlib.use("Qt5Agg")
+    app = QApplication(sys.argv)
+    app.setApplicationName("MNELAB")
+    app.setOrganizationName("cbrnr")
+    main = MainWindow()
+    if len(sys.argv) > 1:  # open files from command line arguments
+        for f in sys.argv[1:]:
+            main.load_file(f)
+    sys.exit(app.exec_())
--- a/mnelab/dialogs/calcdialog.py
+++ b/mnelab/dialogs/calcdialog.py
@@ -1,0 +1,13 @@
+from PyQt5.QtWidgets import QDialog, QLabel, QVBoxLayout, QDialogButtonBox
+
+
+class CalcDialog(QDialog):
+    def __init__(self, parent, title, message):
+        super().__init__(parent)
+        self.setWindowTitle(title)
+        vbox = QVBoxLayout(self)
+        label = QLabel(message)
+        button = QDialogButtonBox(QDialogButtonBox.Cancel)
+        button.rejected.connect(self.close)
+        vbox.addWidget(label)
+        vbox.addWidget(button)
--- a/mnelab/dialogs/runicadialog.py
+++ b/mnelab/dialogs/runicadialog.py
@@ -4,7 +4,7 @@ from PyQt5.QtCore import Qt
 
 
 class RunICADialog(QDialog):
-    def __init__(self, parent, nchan):
+    def __init__(self, parent, nchan, have_picard=True):
         super().__init__(parent)
         self.setWindowTitle("Run ICA")
         vbox = QVBoxLayout(self)
@@ -13,6 +13,8 @@ class RunICADialog(QDialog):
         self.methodbox = QComboBox()
         self.methods = {"Extended Infomax": "extended-infomax",
                         "Infomax": "infomax", "FastICA": "fastica"}
+        if have_picard:
+            self.methods["Picard"] = "picard"
         self.methodbox.addItems(self.methods.keys())
         grid.addWidget(self.methodbox, 0, 1)
         grid.addWidget(QLabel("Number of components:"), 1, 0)
--- a/mnelab/dialogs/runicadialog.py
+++ b/mnelab/dialogs/runicadialog.py
@@ -1,5 +1,5 @@
 from PyQt5.QtWidgets import (QDialog, QVBoxLayout, QGridLayout, QLabel,
-                             QSpinBox, QComboBox, QDialogButtonBox)
+                             QSpinBox, QComboBox, QDialogButtonBox, QCheckBox)
 from PyQt5.QtCore import Qt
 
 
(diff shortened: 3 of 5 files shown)


cbrnr:
@larsoner @agramfort @dengemann do you have an idea? Or do you know someone who might be able to help? Without solving this it will be difficult to continue with the GUI...

cbrnr:
Maybe https://github.com/spyder-ide/spyder/blob/master/spyder/utils/workers.py provides some insights, I will look into it.

larsoner:
Have you looked where the time hangup is? If it's in something like `FastICA.fit(...)` I can't think of a good way to stop it, because that is a call to `sklearn`, and we can't change their code. You could 1) extract the data to fit into a `memmapped` array, 2) spawn a new process with `subprocess.POpen` that uses `ica._fit(...)` which operates on the `ndarray` directly, 3) control / monitor that subprocess as it runs, and 4) save the result (in the second Python process) somewhere and then load it with your first one.

I'd assume that there is a way to do this with `multiprocess.pool` but it sounds like you've tried it and it doesn't work.

cbrnr:
Thanks @larsoner. The time is spent in the ICA function, yes (I mostly tested MNE's own Infomax algorithm). Using memory mapped arrays could be an option, but I'd prefer to call the (external) function directly without additional hacks. I don't think multiprocessing is going to work, but I still hope that multithreading will somehow magically solve this issue. I don't have time to look into that in the near future, but hopefully the Spyder solution is somehow applicable to this problem.

@ccordoba12, do you think that multithreading is going to work?

ccordoba12:
@cbrnr, multithreading or multiprocessing?

cbrnr:
I tried multiprocessing but it doesn't work (see the top post for the reasons), so I thought that multithreading should work hopefully without the extra overhead.

mmagnuski:
> It seems like the other process (the ICA calculation) is trying to change the GUI, which I don't understand at all.

I don't have much experience in multiprocessing but can this be related to the fact that ICA may be using multiprocessing via joblib?

mmagnuski:
BTW - it might be useful to add progress kwarg to longer-running mne functions like ICA. That kwarg could take functions passed by the user (so that the user can get the progress the way they want - for example by updating GUI progressbar). If we consider this it would be then easy to add an option where the output of progress function could cancel execution (False -> cancel, None or True -> continue).
If you have problems with multiproc / threading it maybe enough to just run the ICA via GUI without the ability to stop, at least for now. I am sure I would teach using mnelab once it has ICA and some basic ERP functionality even if there is no possiblity to stop the ICA (you could give a big warning in the ICA GUI stating that user would be able to abandon ICA only via process manager and therefore loose all unsaved progress).

cbrnr:
@mmagnuski yes, making longer-running MNE functions interruptible would likely work around this issue. I'm just not sure if it would be really that straightforward since I'd still need to send a stop signal via the GUI, which in turn has to be responsive. AFAIK even if this can be made to run in a separate thread, there is no method to safely stop a thread, and a new kwarg would solve at least this issue.

Maybe you are right, and I should just go on with MNELAB and think about a proper solution later on. Even though my time is also pretty limited, I'd still like to try one last time to see if something from the Spyder codebase can be used in this case (see the link to their code in a previous comment).

cbrnr:
Yeah, I'm not sure that this will work either and now that you put it like that it really might not work.

So have we encountered the limits of Python 😱? I can't believe it's not possible to do such a seemingly simple thing. I always thought it's because I just don't know enough about that kind of stuff, but we'll see, I need to do more research. I thought of more complicated solutions, e.g. I could spawn a separate Python process for all computations right after starting the main Python process which just handles the GUI. But this would mean there would need to be proper IPC to bounce at least meta-information around. There has to be a simple solution 🤔!

larsoner:
> So have we encountered the limits of Python ￼? I can't believe it's not possible to do such a seemingly simple thing. I always thought it's because I just don't know enough about that kind of stuff, but we'll see, I need to do more research. I thought of more complicated solutions, e.g. I could spawn a separate Python process for all computations right after starting the main Python process which just handles the GUI.

I hate to say this but `TraitsUI` might make things easier here. The coreg GUI spawns a new thread (or process?) to compute the BEM solution (and I think Traits uses one for processing events?) and the rest of the GUI stays usable during that time.

larsoner:
https://github.com/mne-tools/mne-python/blob/master/mne/gui/_coreg_gui.py#L1338

agramfort:
why don't you use joblib?

cbrnr:
I don't think joblib is the right tool for this use case. I looked into it, but it seemed to me that it can't just push a job onto a subprocess without doing any calculations in the main process. Anyway, I think the current solution works quite nicely (but it'd be nice if others could confirm).

larsoner:
On macOS High Sierra `mne` env I did `python mnelab.py`, File->Open `sample_audvis_raw.fif`, then `Tools->Run ICA...` and nothing happens (no window popup, etc.). Doing `Tools->Filter data...` pops up a window asking for freq bounds, if I tell it 1-40 after a few seconds it gives me a second entry in the left list including a filtered verison. So I think ICA is not working for me.

larsoner:
:man_facepalming: nope!

There is no `picard` option, I refuse to try it :)

Cancelling the computation seems to work. Running it does, too, except that I ran it with 100 components (would be nice if the cancel screen mentioned this) and the result said it had 364 components. Maybe this has to do with the issue about what the various processing stages mean?

Unrelated suggestions:
- you can't use a `float` number of components (i.e., explained var proportion)
- options for tolerance would be nice
- STIM is listed before EEG in the channel counts, maybe all data channels should come first?

cbrnr:
Got it 😄 - thanks for the feedback and suggestions, right now I wanted to confirm that the multiprocess business is working. Of course I need to add a lot of parameters to the GUI next. Of course Picard will be added, as will all of your suggestions. I was also wondering how to best select channel types. Right now, the user needs to pick channels via the Edit menu, which creates a new data set with the selected channels, after which ICA can be started. Maybe it would make sense to add a picks selection that doesn't remove channels but only influences ICA computation.

cbrnr:
@larsoner stim channels are now always at the end of the channel type list

PR #19, opened 2019-04-18
Set up Azure pipelines

cbrnr:


azure-pipelines.yml
```diff
@@ -1,28 +1,30 @@
...
+  - task: UsePythonVersion@0
+    inputs:
+      versionSpec: '$(python.version)'
+    displayName: 'Use Python $(python.version)'
+  - script: |
+      python -m pip install --upgrade pip setuptools wheel
+      pip install -r requirements.txt
+    displayName: 'Install dependencies'
+  - script: python setup.py
+      displayName: 'Install MNELAB'
```
larsoner: this has two extra spaces, align it with `script` and you should be good

cbrnr: Thanks, this worked!

cbrnr:
Hmmm, this is strange - this test passes locally. @larsoner do you have any idea?

larsoner:
Install `pytest-faulthandler` so you can hopefully get a useful traceback

cbrnr:
Doesn't look too useful to me...

larsoner:
Topmost:
```
File "/opt/hostedtoolcache/Python/3.6.8/x64/lib/python3.6/site-packages/pytestqt/plugin.py", line 57 in qapp
```
It tells you that it failed somewhere in setting up the app, specifically:

https://github.com/pytest-dev/pytest-qt/blob/master/pytestqt/plugin.py#L57

So there is probably a Qt lib problem. Before running tests, I'd try some simple Python commands to create and start a Qt app, then close it. If these succeed, perhaps some config problem exists with pytest-qt. If these fail, then there is some problem with your PyQt install that makes it not work at all. I'd also throw in something like `python -c "import matplotlib; matplotlib.use('qt5agg'); import matplotlib.pyplot as plt; plt.figure()"` as a command to see if it succeeds.

cbrnr:
Thanks, I didn't even think to check that line. This is very helpful, I will try to solve this tomorrow.

cbrnr:
I don't know. Matplotlib with `qt5agg` backend works. The problem is also not `pytest-qt`. Now the job crashes when I initialize the main window in line 76 (`super.__init__()`). I have no idea why this doesn't work.

larsoner:
Are both you and `pytest-qt` starting a `QApplication` instance? Can this be problematic somehow?

cbrnr:
I have the problem also when `pytest-qt` is not installed.

larsoner:
You could try installing an older version of PyQt (5.11?). Other than that I'm not sure. Have you tested locally on Windows?

cbrnr:
PyQt 5.11 doesn't work either. The test is now just:

```
from PyQt5.QtWidgets import QWidget

def test_pyqt():
    widget = QWidget()
```

It still fails. I use a Linux job on Azure, not Windows. Locally I've only tested this on macOS, and it works. Am I doing something wrong?

cbrnr:
Same error on Windows and macOS jobs.

cbrnr:
Finally this works really nicely on all three platforms. There were two issues:

1. I needed to install `pytest-xvfb`, otherwise creating windows didn't work.
2. On the Linux agent, I needed to apt install `libxkbcommon-x11-0`.

larsoner:
Which thing fixed windows?

cbrnr:
The `pytest-xvfb` module.

larsoner:
Weird, wonder why we don't need it for MNE (and matplotlib, etc. does not need it). Also the naming suggests it's only useful for Linux (since Xvfb is a Linux thing). No wonder it was difficult to figure out it was needed!

PR #22, opened 2019-04-30
[WIP] Support for XDF

cbrnr:
Prepare support for loading XDF files. Currently, XDF can't be loaded yet, but a dialog window shows a list of all available streams contained in the file. Marker streams (streams with fs=0Hz) are disabled. The plan is to load the selected stream (for now, only one stream can be selected and loaded). Marker streams will be imported as annotations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ Optional dependencies provide additional features if installed:
 - [scikit-learn]() (ICA computation via FastICA)
 - [python-picard](https://pierreablin.github.io/picard/) (ICA computation via PICARD)
 - [pyEDFlib](https://github.com/holgern/pyedflib) (export raw to EDF/BDF)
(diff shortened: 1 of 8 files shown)


cbrnr:
A single XDF stream can now be loaded.

Caveats:
- Large files slow down the workflow quite a bit. This is especially visible when plotting a large stream.
- Works best with EEG streams. Other types can be loaded, but plotting an audio stream crashes Python, because MNE is not prepared to plot streams with such a high sampling rate (and therefore some heuristically chosen values cause a buffer overflow). This is e.g. the case for the `sample_xdf.xdf` file in the [XDF repository](https://github.com/sccn/xdf).
- The model and view are not cleanly separated yet. This is problematic when opening an XDF file from the File - Recent menu, which currently crashes MNELAB.

Despite these caveats (I will at least work on the last one), comments are welcome!

PR #32, opened 2019-07-01
DOC: Add conda install instructions, and conda-forge and PyPI badges

hoechenberger:


cbrnr:
Nice - thanks @hoechenberger!

PR #43, opened 2019-07-04
ADD Apply ICA function

### LINKED ISSUE #40: Apply ICA
As mentioned in #34, and way to use ica.apply() would be a nice feature.

vferat:
Fixes #40 @cbrnr & @mmagnuski can you test if if works for you ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,5 +9,3 @@
 ## [0.1.0] - 2019-06-27
 ### Added
 - Initial release (by [Clemens Brunner](https://github.com/cbrnr))
-
-
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [Unreleased] - 2019-04-26
 ### Added
 - Show version number in About dialog ([#28](https://github.com/cbrnr/mnelab/pull/28) by [Clemens Brunner](https://github.com/cbrnr))
+- Add Apply ICA to Tools menu that allows user to apply fitted ica to the current dataset by removing selected components from the signal. ([#43](https://github.com/cbrnr/mnelab/pull/43) by [Victor Férat](https://github.com/vferat))
 
 ### Fixed
 - Fix crash when no channel labels are present in XDF file by [Clemens Brunner](https://github.com/cbrnr))
@@ -8,3 +9,5 @@
 ## [0.1.0] - 2019-06-27
 ### Added
 - Initial release (by [Clemens Brunner](https://github.com/cbrnr))
+
+
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## [Unreleased] - 2019-04-26
 ### Added
 - Show version number in About dialog ([#28](https://github.com/cbrnr/mnelab/pull/28) by [Clemens Brunner](https://github.com/cbrnr))
-- Add Apply ICA to Tools menu that allows user to apply fitted ica to the current dataset by removing selected components from the signal. ([#43](https://github.com/cbrnr/mnelab/pull/43) by [Victor Férat](https://github.com/vferat))
+- Add "Apply ICA" to Tools menu that allows users to apply a fitted ICA solution to the current dataset ([#43](https://github.com/cbrnr/mnelab/pull/43) by [Victor Férat](https://github.com/vferat))
 
 ### Fixed
 - Fix crash when no channel labels are present in XDF file by [Clemens Brunner](https://github.com/cbrnr))
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -194,7 +194,8 @@ class MainWindow(QMainWindow):
                                                            self.find_events)
         self.actions["run_ica"] = tools_menu.addAction("Run &ICA...",
                                                        self.run_ica)
-
+        self.actions["apply_ica"] = tools_menu.addAction("Apply &ICA...",
+                                                         self.apply_ica)
         view_menu = self.menuBar().addMenu("&View")
         self.actions["statusbar"] = view_menu.addAction("Statusbar",
                                                         self._toggle_statusbar)
@@ -510,6 +511,11 @@ class MainWindow(QMainWindow):
             else:
                 self.model.current["ica"] = res.get(timeout=1)
                 self.data_changed()
+
+    def apply_ica(self):
+        """Apply current fitted ICA."""
+        self.auto_duplicate()
+        self.model.apply_ica()
 
     def filter_data(self):
         """Filter data."""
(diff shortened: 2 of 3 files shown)


mnelab/mainwindow.py
```diff
@@ -494,12 +496,21 @@ def run_ica(self):
...
+
             if not dialog.ortho.isHidden():
                 fit_params["ortho"] = dialog.ortho.isChecked()
+                self.model.history.append('fit_params["ortho"] = {}'.format(dialog.ortho.isChecked()))
+
             ica = mne.preprocessing.ICA(method=dialog.methods[method],
                                         fit_params=fit_params)
+            self.model.history.append("ica = mne.preprocessing.ICA("
+                                    + "method={}".format(dialog.methods[method])
+                                    + " ,fit_params=fit_params)")
```
mmagnuski: space should be after, not before coma

mmagnuski: btw, you could also do:
```
("text text text "
 "text text text")
```
instead of 
```
"text text text "
 + "text text text"
```

cbrnr: Also, feel free to use f-strings!

CHANGELOG.md
```diff
@@ -1,10 +1,13 @@
 ## [Unreleased] - 2019-04-26
 ### Added
 - Show version number in About dialog ([#28](https://github.com/cbrnr/mnelab/pull/28) by [Clemens Brunner](https://github.com/cbrnr))
+- Add Apply ICA to Tools menu that allows user to apply fitted ica to the current dataset by removing selected components from the signal. ([#43](https://github.com/cbrnr/mnelab/pull/43) by [Victor Férat](https://github.com/vferat))
```
mmagnuski: you should ad the change to a new version (not past unreleased), mnelab is already after `0.1.0`

cbrnr: That's OK, unreleased means "currently unreleased", so this is going to be the next release.

CHANGELOG.md
```diff
@@ -1,10 +1,13 @@
...
 
 ### Fixed
 - Fix crash when no channel labels are present in XDF file by [Clemens Brunner](https://github.com/cbrnr))
 
 ## [0.1.0] - 2019-06-27
 ### Added
 - Initial release (by [Clemens Brunner](https://github.com/cbrnr))
+
+
```
cbrnr: Remove 1 empty line please

mnelab/mainwindow.py
```diff
@@ -494,12 +496,23 @@ def run_ica(self):
             method = dialog.method.currentText()
             exclude_bad_segments = dialog.exclude_bad_segments.isChecked()
             fit_params = {}
+            self.model.history.append("fit_params = {}")
```
cbrnr: Instead of extending an empty dict, you could add the finished `fit_params` to the history after everything has been constructed.

mnelab/mainwindow.py
```diff
@@ -494,12 +496,23 @@ def run_ica(self):
...
+
             if not dialog.extended.isHidden():
                 fit_params["extended"] = dialog.extended.isChecked()
+                self.model.history.append('fit_params["extended"] = %s'
+                                          % (dialog.extended.isChecked()))
```
cbrnr: Please use f-strings everywhere (instead of `%` or `.format`)

CHANGELOG.md
cbrnr: Minor nitpick: please rephrase to:
```
- Add "Apply ICA" to Tools menu that allows users to apply a fitted ICA solution to the current dataset ([#43](https://github.com/cbrnr/mnelab/pull/43) by [Victor Férat](https://github.com/vferat))

File: mnelab/mainwindow.py [hunk 6]
```diff
@@ -494,12 +496,18 @@ def run_ica(self):
...
+
             if not dialog.extended.isHidden():
                 fit_params["extended"] = dialog.extended.isChecked()
+
             if not dialog.ortho.isHidden():
                 fit_params["ortho"] = dialog.ortho.isChecked()
+
             ica = mne.preprocessing.ICA(method=dialog.methods[method],
                                         fit_params=fit_params)
+            self.model.history.append("ica = mne.preprocessing.ICA("
```
cbrnr: It is better if all string parts are f-strings

mnelab/mainwindow.py
```diff
@@ -509,8 +517,16 @@ def run_ica(self):
                 pool.terminate()
             else:
                 self.model.current["ica"] = res.get(timeout=1)
+                self.model.history.append("ica.fit(inst=raw, "
```
cbrnr: Same here, please make all lines f-strings

mnelab/mainwindow.py
```diff
@@ -194,7 +194,8 @@ def __init__(self, model):
                                                            self.find_events)
         self.actions["run_ica"] = tools_menu.addAction("Run &ICA...",
                                                        self.run_ica)
-
+        self.actions["apply_ica"] = tools_menu.addAction("Apply &ICA...",
```
cbrnr: Remove the "..." (which is only used if a dialog window will pop up)

mnelab/mainwindow.py
```diff
@@ -194,7 +194,8 @@ def __init__(self, model):
                                                            self.find_events)
         self.actions["run_ica"] = tools_menu.addAction("Run &ICA...",
```
cbrnr: Can you insert a separator before the two ICA actions?

cbrnr:
Currently, you can only mark bad ICs with "Plot ICA components", which requires a montage. It would be nice if you could mark bad ICs also with the "Plot ICA sources" by clicking on the IC label or IC trace (just like it is possible with Plot raw). This is something MNE should implement and not MNELAB, so I'm going to open an issue there.

vferat:
> 
> 
> Currently, you can only mark bad ICs with "Plot ICA components", which requires a montage. It would be nice if you could mark bad ICs also with the "Plot ICA sources" by clicking on the IC label or IC trace (just like it is possible with Plot raw). This is something MNE should implement and not MNELAB, so I'm going to open an issue there.

You can mark ICs with "Plot ICA sources" by clicking on the IC trace,  in my case , it becomes red and is rejected when applying ICA. After applying, the source is set to 0

cbrnr:
OK, true, but it is still not consistent because clicking on the IC label should also mark it bad. Currently, it tries to open a topomap, which is not possible if there is no montage.

vferat:
Yes I agree . As you said, the issue should be resolve in MNE repo.
The problem is to be consistent both with "raw.plot" and "ica_plot_components"  in a way that can easly displays  topomaps if available

cbrnr:
Thanks @vferat!

CHANGELOG.md
mmagnuski: Ah, sorry, I read it wrong.

mmagnuski: the dates are just a bit confusing, as this unreleased section has earlier date than the `0.1.0` release so I read it as past unreleased changes.

cbrnr: Yes, I've changed the date to XX.XX.XXXX to avoid confusion.

PR #50, opened 2019-07-09
Create events from annotations

### LINKED ISSUE #47: Allow conversion of annotations to events (required for BrainVision)
<strike>In the data recorded using our BrainProducts system, there is no separate "stim channel", and I extract the events via
```python
mne.find_events(raw_data)
```
This is not possible using MNELAB, as the "Find events..." dialog insists I select a "stim channel", which doesn't make sense in my case. There should be the option to leave the stim channel blank, in which case `mne.find_events()` would be called with `stim_channel=None`.</strike>

I'm sorry, the original issue description doesn't make sense anymore with recent versions of `mne-python`. Here's an updated, corrected description:

The events of BrainVision data have to be generated from the annotations of the raw data:
```python
import mne

data = mne.io.read_raw_brainvision('eeg_recording.vhdr')
events = mne.events_from_annotations(data)
```
This work flow is currently not supported in MNELAB.

cbrnr:
Fixes #47. @hoechenberger please check if this solves the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Show version number in About dialog ([#28](https://github.com/cbrnr/mnelab/pull/28) by [Clemens Brunner](https://github.com/cbrnr))
 - Add "Apply ICA" to Tools menu that allows users to apply a fitted ICA solution to the current dataset ([#43](https://github.com/cbrnr/mnelab/pull/43) by [Victor Férat](https://github.com/vferat))
 - Plot raw shows the dataset name in the title (instead of "Raw data") (by [Clemens Brunner](https://github.com/cbrnr))
+- Add option to convert annotations to events ([#50](https://github.com/cbrnr/mnelab/pull/50) by [Clemens Brunner](https://github.com/cbrnr))
 
 ### Fixed
 - Fix crash when no channel labels are present in XDF file (by [Clemens Brunner](https://github.com/cbrnr))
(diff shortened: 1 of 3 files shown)


hoechenberger:
@cbrnr _In principle_,  this works. However, the events have unexpected names, such as `10001`, `10002` etc.

The names should be based on the annotation descriptions instead:
```python
events = mne.events_from_annotations(data)
Used Annotations descriptions: ['Event/100', 'Event/110', 'Event/120', 'Event/140', 'Event/160', 'Event/180', 'Event/200', 'Event/230', 'New Segment/']
events[1]
Out[5]: 
{'Event/100': 10001,
 'Event/110': 10002,
 'Event/120': 10003,
 'Event/140': 10004,
 'Event/160': 10005,
 'Event/180': 10006,
 'Event/200': 10007,
 'Event/230': 10008,
 'New Segment/': 99999}
```

(`100`, `110` etc are the event marker codes we used)

cbrnr:
This is an MNE issue, and I think it was fixed in https://github.com/mne-tools/mne-python/issues/6267 - can you check if it works with the current MNE master?

hoechenberger:
Ugh, I hadn't followed this development. No, this is not fixed -- apparently the behavior I reported above is what was introduced in the PR you referenced. I have no idea why they did it this way. Seems very unpractical to me.

hoechenberger:
Since this seems to be a general issue with MNE, I'm probably going to open an issue there.

cbrnr:
Yes, please refer to this issue here.

hoechenberger:
I think it's an incompatibility / bug in the BrainVision event parser. Will try to create an error description and file a bug report with MNE. I believe you can already merge this PR in the meantime, as convertig annotations to events _did_ work, it's just that the event names are currently off for me with BrainVision data.

Issue #53, opened 2019-07-11
Strange text is printed when epoching error dialog is shown


cbrnr:
When I supply invalid epoching values (e.g. tmin > tmax) in the dialog, an error dialog is shown. However, a bunch of strange unreadable text is also printed in the main window:

<img width="950" alt="Screen Shot 2019-07-11 at 09 52 52" src="https://user-images.githubusercontent.com/4377312/61032511-dccdc800-a3c1-11e9-86de-2b2a77c02ac4.png">

I have no idea why this is happening and I'd appreciate any help.

cbrnr:
I think this has something to do with the message box starting its own event loop (i.e. it is called with the `exec_` method). If I instead show the message box with `open`, I don't get the strange text in the main window. However, I can't use the static `QMessageBox.critical` then, so I have to use a custom `QMessageBox` instead, which is OK. The other downside of using `open` is that the dialog slides down as a sheet on macOS, but I'd like to have a proper dialog window instead.

cbrnr:
PS: Interestingly, the [Qt documentation[(https://doc.qt.io/qt-5/qdialog.html#exec) recommends not to use `exec_`.

PR #55, opened 2019-07-16
ADD Interpolate bad Channels

vferat:
This PR add the ability to use `.interpolate_bads` on raw, epoch and evoked data


cbrnr:
Thanks, I'll take a look. Just to make life easier in the future, could you please create PRs from branches other than master?

vferat:
No problem, sorry for the mess, I was not aware about issue due to PRs from master. I didn't change anything more than adding ` .interpolate_bads`. Thanks for the advice

cbrnr:
No big deal, it just requires me to type 2 lines instead of 1 when checking out your branch 😄. I'll push a few more changes and this will be ready to go.

cbrnr:
I'm catching a ValueError if some channels do not have location information. This works, but it has the same visual glitch as described in #53. So if anyone knows how to fix this this would be absolutely great!

cbrnr:
I'm catching a ValueError if some channels do not have location information. This works, but it has the same visual glitch as described in #53. So if anyone knows how to fix this this would be absolutely great!

cbrnr:
Thanks @vferat!

PR #59, opened 2019-08-12
Add history for changes in channel properties dialog

### LINKED ISSUE #56: History when plotting and using channel properties
There is no history for actions done when plotting data (such as setting bads channels) either when using the "channel properties" dialog.

cbrnr:
Fixes #56. I don't know how to infer changes to bad channels made in the interactive figure window though. Ideas @vferat @mmagnuski @larsoner?


cbrnr:
OK, I found a solution :smile:.

PR #76, opened 2019-09-03
Add version numbers for dependencies (instead of just True/False)

cbrnr:
This extends the dict by setting the values to the package versions if they exist - if a package does not exist, the value remains `False`, e.g.:

```
{'PyQt5': '5.13.0',
 'matplotlib': '3.1.1',
 'mne': '0.19.dev0',
 'numpy': '1.17.1',
 'picard': False,
 'pybv': '0.2.0',
 'pyedflib': '0.1.14',
 'pyxdf': '1.15.2',
 'scipy': '1.3.1',
 'sklearn': '0.21.3'}
```


cbrnr:
Note: This makes version comparisons possible. The recommended way is to use `packaging` (available via `pip`, but will be included in Python 3.8's standard library):

```
from packaging import version
version.parse("2.3.1") < version.parse("10.1.2")
```

PR #107, opened 2019-12-13
Fix montage

cbrnr:
MNE now uses `mne.channels.make_standard_montage` instead of `mne.channels.read_montage`.


cbrnr:
I guess we'll need to require MNE >= 0.19 for this to work.

PR #108, opened 2020-01-15
Add option to append data sets

### LINKED ISSUE #70: Add option to append data sets
This could be a dialog showing all currently opened raw objects, so users first need to open all desired files, which they can then concatenate in the dialog window. This creates a new dataset.

stralu:
Fixes #70.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 ## [Unreleased] - XXXX-XX-XX
+### Added
+- Add support for appending data sets ([#108](https://github.com/cbrnr/mnelab/pull/108) by [Lukas Stranger](https://github.com/stralu) and [Clemens Brunner](https://github.com/cbrnr))
+
 ### Changed
 - Use [QtPy](https://github.com/spyder-ide/qtpy) to support both PyQt5 and PySide2 ([#118](https://github.com/cbrnr/mnelab/pull/118) by [Clemens Brunner](https://github.com/cbrnr))
 - Remove resource file and include icons directly ([#125](https://github.com/cbrnr/mnelab/pull/125) by [Clemens Brunner](https://github.com/cbrnr)) 
--- a/mnelab/dialogs/__init__.py
+++ b/mnelab/dialogs/__init__.py
@@ -6,6 +6,7 @@ from .annotationsdialog import AnnotationsDialog
 from .calcdialog import CalcDialog
 from .channelpropertiesdialog import ChannelPropertiesDialog
 from .cropdialog import CropDialog
+from .concatenatedatadialog import ConcatenateDataDialog
 from .epochdialog import EpochDialog
 from .errormessagebox import ErrorMessageBox
 from .eventsdialog import EventsDialog
--- a/mnelab/dialogs/__init__.py
+++ b/mnelab/dialogs/__init__.py
@@ -6,7 +6,7 @@ from .annotationsdialog import AnnotationsDialog
 from .calcdialog import CalcDialog
 from .channelpropertiesdialog import ChannelPropertiesDialog
 from .cropdialog import CropDialog
-from .concatenatedatadialog import ConcatenateDataDialog
+from .appenddialog import AppendDialog
 from .epochdialog import EpochDialog
 from .errormessagebox import ErrorMessageBox
 from .eventsdialog import EventsDialog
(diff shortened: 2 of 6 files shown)


cbrnr:
Thanks @stralu. I have the following suggestions:

1. I would use a fixed name such as `<original_name>_concat` - this eliminates one source of error.
2. Maybe we should only allow to concatenate 1 data set, because otherwise the order of the entries in the list is important. Currently, the list cannot be sorted, so users cannot change this. Alternatively, a nice solution would be if users were able to sort the list via drag and drop.
3. I received when concatenating a data set which was in the list. Maybe we could use `mne.io.base._check_raw_compatibility` to find out if data sets are compatible?
4. It seems like currently it is only possible to concatenate data files that are associated with a file. Would it be practical to be able to concatenate also processed data sets (e.g. a filtered data set)? Currently, these don't show up in the list.

cbrnr:
Maybe one idea to implement this is to have two lists in the dialog window. The left list shows all compatible data sets, whereas the right list contains the selected data sets that get concatenated to the current data set. Then only the right list must be sortable via drag and drop.

stralu:
Thanks @cbrnr for your comments!

I like 1,3 and 4 and will go for them. But, I don't think that we need to implement the drag and drop list for ordering the data. Hardly any data set will fit neatly together (sample-by-sample) and thus the order of concatenation will not be of advantage for detrending/filtering the concatenated data set. Neither, is the order of advantage whenever the data is epoched and averaged in a later processing step. Or do I miss a point?

cbrnr:
I thought that if you recorded e.g. 5 runs in 5 separate files it might be convenient to concatenate them in the correct (temporal) order.

cbrnr:
Thanks for the update, I have rebased to master and pushed some additional changes. This feature looks almost ready, the following things should be addressed before this will be merged:

1. The action should be disabled if there is no compatible data set to append. Currently, the dialog is enabled even if only one data set is loaded (but of course both lists are empty).
2. The history entry is not quite correct yet. If you take a look at which entries are generated, you'll see that `data` is the currently active data set, and `datasets` is a list containing all open data sets. Currently, `data.append` uses a list of names as the argument, but it should be a list of references to the `dataset` entries (with the desired indices). For example, `data.append(['S001R02', 'S001R03'])` is not correct because the list entries should be `raw` objects, so it should say `data.append([datasets[1|, datasets[2])` (assuming that we append data sets with indices 1 and 2).
3. I'm not sure if the selected order is correctly maintained. I have tested this quickly and it seems like the order of the appended data sets does not correspond to the list order.
4. Both issues 2 and 3 could be fixed if you didn't use the names but the indices of the data sets to actually append the data in the model.
5. Maybe we should rename this action to "Append data..." (including the method name).
6. It would be nice if multiple rows could be selected and dragged over (not a must if the implementation is too difficult though).

You could use data from https://physionet.org/content/eegmmidb/1.0.0/ for testing, these are all compatible and can be appended.

cbrnr:
Nice! Would it be possible to rebase your work to my latest commit (i.e. 42c052f) and then force-push? That way, your renaming change in 6d2d281 isn't required anymore, nor is the last merge commit 8eb9f3f. Let me know if I should do this or if you want to go for it.

cbrnr:
Scratch that, I see your changes were not identical and the last merge commit has my changes. So no need to rebase! I'll quickly test it now.

cbrnr:
But still, your changes don't have 516f336 - I removed the custom list widget because it's not really needed.

cbrnr:
I think I just messed this up. Can you please *not* pull from this branch?

cbrnr:
All good, I think I fixed it!

cbrnr:
I think the correct history entry will be difficult with the current data model (see #127), so let's not do this in this PR. This leaves the following two points to be checked:

1. I'm not sure if the selected order is correctly maintained. I have tested this quickly and it seems like the order of the appended data sets does not correspond to the list order.
2. Data sets without a file (e.g. as created when filtering data) do not show up in the list (but they should if they are compatible).

cbrnr:
Alright, data sets without a file already work, I just tested with a filtered data set, which is not compatible because we assert that both lowpass and highpass filters must be identical.

cbrnr:
The order seems to be correct. I will do another (final) round of checks and hopefully this will be ready to be merged!

cbrnr:
Thanks @stralu!

PR #109, opened 2020-01-25
Add more items to history

cbrnr:
- [x] Initialize with `import mne`
- [x] Show `data.plot(events=events)` if events are available
- [x] History for `plot_psd()`
- [x] Include switching and/or creating new data sets in history
- [x] History for setting montage


cbrnr:
Adding history for editing events and/or annotations requires more effort, I'll do that in a separate PR.

PR #119, opened 2020-01-30
ENH: Always run via framework build on macOS conda

hoechenberger:
This PR ensures that we're always running MNELAB with a "framework build" of Python on macOS, if a `conda`-installed Python interpreter is used.

The way this works is the following:
- first, we check whether we're running macOS and in a `conda` environment (upon environment activation, the `CONDA_PREFIX` env variable is set)
- if that's not the case, run the application as usual
- otherwise, spawn a new `pythonw` process (i.e., a framework build) to execute `__main__.py` again, preserving all command line arguments that were initially passed, and set an environment variable `MNELAB_RUNNING_PYTHONW`, indicating we're now using the framework build
- now, when the newly spawned Python interpreter executes `__main__.py`, it will find that we're using a framework build, because the environment variable is set; and will proceed to start the application in the "usual" manner

We have to jump through these hoops bc there's no built-in way to determine whether Python was started as `pythonw`, so we introduce an environment variable and _always_ spawn a subprocess on macOS if we're using `conda`.

With this, both the `mnelab` entry point and `python -m mnelab` start up the application as the users would expect.

### DIFF
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -3,6 +3,8 @@
 # License: BSD (3-clause)
 
 import multiprocessing as mp
+mp.set_start_method("spawn", force=True)  # required for Linux/macOS
+import pebble
 from sys import version_info
 import traceback
 from os.path import isfile, isdir
@@ -88,9 +90,6 @@ class MainWindow(QMainWindow):
             sets. This decouples the GUI from the data (model/view).
         """
         super().__init__()
-
-        mp.set_start_method("spawn", force=True)  # required for Linux/macOS
-
         self.model = model  # data model
         self.setWindowTitle("MNELAB")
 
@@ -635,19 +634,26 @@ class MainWindow(QMainWindow):
             self.model.history.append(f"ica = mne.preprocessing.ICA("
                                       f"method={dialog.methods[method]}, "
                                       f"fit_params={fit_params})")
-            pool = mp.Pool(1)
+
             kwds = {"reject_by_annotation": exclude_bad_segments}
-            res = pool.apply_async(func=ica.fit,
-                                   args=(self.model.current["data"],),
-                                   kwds=kwds, callback=lambda x: calc.accept())
+            print(mp.get_start_method)
+            pool = pebble.ProcessPool(max_workers=1)
+            process = pool.schedule(function=ica.fit,
+                                    args=(self.model.current["data"],),
+                                    kwargs=kwds)
+            process.add_done_callback(lambda x: calc.accept())
+            pool.close()
+
             if not calc.exec_():
-                pool.terminate()
+                pool.stop()
+                pool.join()
             else:
-                self.model.current["ica"] = res.get(timeout=1)
+                self.model.current["ica"] = process.result()
                 self.model.history.append(f"ica.fit(inst=raw, "
                                           f"reject_by_annotation="
                                           f"{exclude_bad_segments})")
                 self.data_changed()
+                pool.join()
 
     def apply_ica(self):
         """Apply current fitted ICA."""
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -2,7 +2,6 @@
 #
 # License: BSD (3-clause)
 
-import multiprocessing as mp
 import pebble
 from sys import version_info
 import traceback
@@ -38,7 +37,6 @@ import mnelab.resources  # noqa
 __version__ = "0.6.0.dev0"
 
 MAX_RECENT = 6  # maximum number of recent files
-mp.set_start_method("spawn", force=True)  # required for Linux/macOS
 
 
 def read_settings():
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -3,7 +3,6 @@
 # License: BSD (3-clause)
 
 import multiprocessing as mp
-mp.set_start_method("spawn", force=True)  # required for Linux/macOS
 import pebble
 from sys import version_info
 import traceback
@@ -39,6 +38,7 @@ import mnelab.resources  # noqa
 __version__ = "0.6.0.dev0"
 
 MAX_RECENT = 6  # maximum number of recent files
+mp.set_start_method("spawn", force=True)  # required for Linux/macOS
 
 
 def read_settings():
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -636,7 +636,6 @@ class MainWindow(QMainWindow):
                                       f"fit_params={fit_params})")
 
             kwds = {"reject_by_annotation": exclude_bad_segments}
-            print(mp.get_start_method)
             pool = pebble.ProcessPool(max_workers=1)
             process = pool.schedule(function=ica.fit,
                                     args=(self.model.current["data"],),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,12 @@
 ### Fixed
 - Fix blurry icons on macOS HiDPI screens ([#102](https://github.com/cbrnr/mnelab/pull/102) by [Clemens Brunner](https://github.com/cbrnr))
 - Use `mne.channels.make_standard_montage` instead of deprecated `mne.channels.read_montage` ([#107](https://github.com/cbrnr/mnelab/pull/107) by [Clemens Brunner](https://github.com/cbrnr))
+- Ensure MNELAB is run using a "framework build" of Python on `conda` installations on macOS ([#119](https://github.com/cbrnr/mnelab/pull/119) by [Richard Höchenberger](https://github.com/hoechenberger))
 
 ### Changed
 - Use environment markers in `setup.py` for `install_requires` ([#105](https://github.com/cbrnr/mnelab/pull/105) by [Clemens Brunner](https://github.com/cbrnr))
 - Bump required minimum MNE version to 0.19 ([#107](https://github.com/cbrnr/mnelab/pull/107) by [Clemens Brunner](https://github.com/cbrnr))
+- Spawn ICA process pool via `Pebble` instead of `multiprocessing` to avoid Python segfaulting on macOS `conda` installations ([#119](https://github.com/cbrnr/mnelab/pull/119) by [Richard Höchenberger](https://github.com/hoechenberger))
 
 ## [0.5.2] - 2019-10-30
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ MNELAB requires Python >= 3.6. In addition, the following Python packages are re
 - [PyQt5](https://www.riverbankcomputing.com/software/pyqt/download5) >= 5.10.0
 - [numpy](http://www.numpy.org/) >= 1.14.0
 - [scipy](https://www.scipy.org/scipylib/index.html) >= 1.0.0
+- [Pebble](https://pebble.readthedocs.io/)
 - [matplotlib](https://matplotlib.org/) >= 2.0.0
 - [mne](https://github.com/mne-tools/mne-python) >= 0.19.0
 - [pyobjc-framework-Cocoa](https://pyobjc.readthedocs.io/en/latest/) >= 5.2.0 (macOS only)
--- a/mnelab/__main__.py
+++ b/mnelab/__main__.py
@@ -3,6 +3,7 @@
 # License: BSD (3-clause)
 
 import sys
+import os
 import matplotlib
 from PyQt5.QtWidgets import QApplication
 from PyQt5.QtCore import Qt
@@ -10,7 +11,7 @@ from PyQt5.QtCore import Qt
 from mnelab import MainWindow, Model
 
 
-def main():
+def _run():
     app_name = "MNELAB"
     if sys.platform.startswith("darwin"):
         try:  # set bundle name on macOS (app name shown in the menu bar)
@@ -37,6 +38,36 @@ def main():
             model.load(f)
     model.view.show()
     sys.exit(app.exec_())
+
+
+def _run_pythonw():
+    """Execute this script again through pythonw.
+
+    This ensures we're using a framework build of Python on macOS.
+    """
+    cmd = [sys.executable + "w", __file__]  # pythonw executable
+
+    # Append command line arguments.
+    if len(sys.argv) > 1:
+        cmd.append(*sys.argv[1:])
+
+    env = os.environ.copy()
+    env["MNELAB_RUNNING_PYTHONW"] = "True"
+
+    import subprocess
+    subprocess.run(cmd, env=env)
+    sys.exit()
+
+
+def main():
+    # Ensure we're always using a "framework build" on macOS.
+    _MACOS_CONDA = sys.platform == "darwin" and "CONDA_PREFIX" in os.environ
+    _RUNNING_PYTHONW = "MNELAB_RUNNING_PYTHONW" in os.environ
+
+    if _MACOS_CONDA and not _RUNNING_PYTHONW:
+        _run_pythonw()
+    else:
+        _run()
 
 
 if __name__ == "__main__":
--- a/mnelab/__main__.py
+++ b/mnelab/__main__.py
@@ -49,7 +49,7 @@ def _run_pythonw():
     import subprocess
 
     cwd = pathlib.Path.cwd()
-    python_path = pathlib.Path(sys.executable).parent / 'pythonw'
+    python_path = pathlib.Path(sys.exec_prefix) / 'pythonw'
 
     if not python_path.exists():
         msg = ('pythonw executable not found. '
--- a/mnelab/__main__.py
+++ b/mnelab/__main__.py
@@ -6,7 +6,7 @@ import sys
 import os
 import matplotlib
 from PyQt5.QtWidgets import QApplication
-from PyQt5.QtCore import Qt, QProcess, QProcessEnvironment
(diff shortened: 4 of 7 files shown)


cbrnr:
Thanks for coming up with a workaround! However, currently this doesn't work for me because of two issues.

First, I get the following error because the module `mnelab` is not found:

```
$ python -m mnelab 
  File "/Users/clemens/Repositories/mnelab/mnelab/__main__.py", line 11, in <module>
    from mnelab import MainWindow, Model
ModuleNotFoundError: No module named 'mnelab'
```

I'm in the correct folder and the same command works with Homebrew Python.

Second, I usually use `python3` instead of `python`, which is why I get another error because only `pythonw` exists but not `python3w`:

```
FileNotFoundError: [Errno 2] No such file or directory: '/Users/clemens/anaconda/bin/python3w': '/Users/clemens/anaconda/bin/python3w'
```

Finally, I have a concern which I couldn't test so far: for the ICA calculation I'm spawning a new process because I need to keep the GUI responsive. I have the feeling that this workaround could break this, but this can easily be tested once at least the first issue is resolved.

hoechenberger:
> ```
> $ python -m mnelab 
> Traceback (most recent call last):
>   File "/Users/clemens/Repositories/mnelab/mnelab/__main__.py", line 11, in <module>
>     from mnelab import MainWindow, Model
> ModuleNotFoundError: No module named 'mnelab'
> ```
> 
> 
> I'm in the correct folder and the same command works with Homebrew Python.

Oh you mean you tried to run this without prior installation of `mnelab`, just from the `mnelab` source directory? Is that really "a thing"? :)

> Second, I usually use `python3` instead of `python`, which is why I get another error because only `pythonw` exists but not `python3w`:

But you don't commonly use `conda`, so this should be a non-issue, actually, no? Could of course add handling for this case.

hoechenberger:
> Oh you mean you tried to run this without prior installation of `mnelab`, just from the `mnelab` source directory? Is that really "a thing"? :)


Should be easy to fix nevertheless. Working on something right now.

cbrnr:
> Oh you mean you tried to run this without prior installation of mnelab, just from the mnelab source directory? Is that really "a thing"? :)

I think this has been working for quite a while, I'm always too lazy to install a package.

> But you don't commonly use `conda`, so this should be a non-issue, actually, no? Could of course add handling for this case.

True, I basically never use `conda`, but maybe others who use it might be writing `python3` as per the official recommendation in [PEP-394](https://www.python.org/dev/peps/pep-0394/).

hoechenberger:
@cbrnr I think I have addressed the `python -m mnelab` execution from the source dir (i.e., without installation); and the invocation via `python3` is working now too.

cbrnr:
Nice, that indeed took care of both issues. However, as I was already expecting there is a problem with subprocesses now. MNELAB segfaults whenever I run ICA:

```
/Users/clemens/anaconda/bin/pythonw: line 3:  4505 Segmentation fault: 11  /Users/clemens/anaconda/python.app/Contents/MacOS/python "$@"
```

The segfault occurs either when clicking on "Cancel" while ICA is running, or after exiting MNELAB after an ICA has been successfully completed. I'm not sure if there's an easy solution to this problem.

hoechenberger:
Any idea what might be the cause of this? So it's related to the subprocess (running MNELAB) spawning new subprocesses (for running ICA)?

cbrnr:
Yes, that's what's probably causing it, but I have no idea why and how to fix it.

I'm thinking if leaving everything as is and just including a statement in README about the focus issue on macOS/conda (and how to resolve it by tab switching) might be the best option. As nice as conda is, but I definitely can't recommend it for GUIs (Qt, wx, ...) on macOS because of this long-standing bug.

hoechenberger:
@cbrnr it seems to be the call to `multiprocessing.Pool.terminate()` that somehow kills the parent process too.

hoechenberger:
I switched from using `subprocess.run()` to `PyQt5.QtCore.QProcess`, and this doesn't produce a segfault anymore: the application still terminates when the multiprocessing pool is stopped upon clicking "Cancel" during the ICA, but it shuts down in a graceful way. Now the question is, why is it shutting down at all :)

cbrnr:
So you're saying we should switch from `multiprocessing` to `QProcess`?

hoechenberger:
> So you're saying we should switch from `multiprocessing` to `QProcess`?

Maybe, idk.

I just switched back to the `master` branch and ran `pythonw -m mnelab`. It crashes with the same error message. So the problem was not introduced by the changes proposed in this PR, and it must be something different… 
I think there's no equivalent to `multiprocessing.Pool` in Qt, or is there? `QProcess` is just individual workers, not entire pools? Only "pool" thing I could find is `QThreadPool`.

cbrnr:
> I just switched back to the master branch and ran `pythonw -m mnelab`. It crashes with the same error message. So the problem was not introduced by the changes proposed in this PR, and it must be something different…

You're right, this must be due to `pythonw`. So currently, running MNELAB with conda on macOS either means:

- `python -m mnelab`: there are glitches with the window manager (menubar focus available only after tab switching back and forth, hovering over grayed out icons in the toolbar makes them black), but multiprocessing works.
- `pythonw -m mnelab`: Multiprocessing produces a segfault, but the GUI works.

> I think there's no equivalent to `multiprocessing.Pool` in Qt, or is there? `QProcess` is just individual workers, not entire pools? The only "pool" thing I could find is `QThreadPool`.

Yeah, I think so too. Maybe this can be pulled off with a single `QProcess` though, because MNELAB spawns only one process during ICA calculation. It's been a while since I've toyed around with that, but I'm pretty sure I've tried and found some problem with it.

hoechenberger:
We could also consider using  a Qt thread then, I believe, and either run the multiprocessing.Pool inside that thread, or run the ICA inside that thread directly?

--

hoechenberger:
In any case we should probably try to create a MWE first... 

--

cbrnr:
Not sure it's worth the effort because IMO this is a bug in conda. There's a long and heated discussion in https://github.com/ContinuumIO/anaconda-issues/issues/199, and what it boils down to is that the official Python build for macOS works because it does the right thing, whereas having separate python/pythonw commands in conda is a hack. No one has neither time nor expertise to work on that, most likely because Anaconda users don't use conda to write GUIs.

hoechenberger:
Just for reference, I suspected that maybe homebrew may have interfered here in some weird way, but I uninstalled homebrew and the issue persists. This is really, really frustrating...

@cbrnr Would you mind keeping this PR open anyway? I'm planning to keep on investigating the issue for a little longer, and would post updates once in a while.

cbrnr:
> Just for reference, I suspected that maybe homebrew may have interfered here in some weird way, but I uninstalled homebrew and the issue persists. This is really, really frustrating...

Quite the opposite as Python is properly packaged in Homebrew I'd say.

> @cbrnr Would you mind keeping this PR open anyway? I'm planning to keep on investigating the issue for a little longer, and would post updates once in a while.

Sure thing. Meanwhile, I think I'll add a warning that I can't recommend conda on macOS for running MNELAB so people should go ahead and use Homebrew.

hoechenberger:
MWE to reproduce the issue:

```python
import sys
import multiprocessing
from PyQt5.QtWidgets import QWidget, QApplication, QPushButton


class Example(QWidget):
    def __init__(self):
        super().__init__()
        button = QPushButton('Push', self)
        button.clicked.connect(self.run_multiprocessing)
        self.show()

    @staticmethod
    def f():
        print('Executing f()...')
        return 1

    def callback(self, x):
        print('...Terminating pool...')
        self.pool.terminate()
        print('...succeess.')

    def run_multiprocessing(self):
        self.pool = multiprocessing.Pool(1)
        self.pool.apply_async(func=self.f,
                              callback=self.callback)

if __name__ == '__main__':
    app = QApplication(sys.argv)
    ex = Example()
    sys.exit(app.exec_())
```

Interestingly, one can push the button (and terminate the `multiprocessing.Pool` exactly once; the second attempt will cause a `Segmentation fault: 11`.

hoechenberger:
If I delete the pool after termination, it doesn't crash anymore.

hoechenberger:
More minimal MWE – show's that the problem isn't bound to PyQt:

```python
import multiprocessing
import time


def f():
    print('Executing f()...')
    return None


if __name__ == '__main__':
    pool = multiprocessing.Pool(1)
    pool.apply_async(func=f)

    time.sleep(1)  # Allow worker to run.

    pool.terminate()
    pool.join()

    print("Deleting join()'ed Pool...")
    del pool
```

```
(mnelab-dev) [hoechenberger:~/Development/qt_test]$ python test2.py 
Executing f()...
Deleting join()'ed Pool...
(mnelab-dev) [hoechenberger:~/Development/qt_test]$ pythonw test2.py 
Executing f()...
Deleting join()'ed Pool...
/Users/hoechenberger/miniconda3/envs/mnelab-dev/bin/pythonw: line 3: 84484 Segmentation fault: 11  /Users/hoechenberger/miniconda3/envs/mnelab-dev/python.app/Contents/MacOS/python "$@"
(mnelab-dev) [hoechenberger:~/Development/qt_test]$ 
```

hoechenberger:
@cbrnr Do you think that's a suitable question for Stack Overflow?

hoechenberger:
I tried setting the executable to `pythonw` via `multiprocessing.set_executable()`, but it seems this is not respected when creating a `multiprocessing.Pool` (even if I pass garbage to `set_executable()`, the worker runs correctly)

cbrnr:
> Do you think that's a suitable question for Stack Overflow?

Definitely, please post the URL if you put it on SO.

Is the problem specific to using a pool? Is there an alternative to a pool (because in our case we know that we only need one subprocess anyway)? The start method has to be "spawn" otherwise Qt gets confused and will eventually crash. Therefore, I added the line `mp.set_start_method("spawn", force=True)` in `mainwindow.py`.

cbrnr:
Also, I think that in addition to posting that on SO this should probably be reported to Anaconda.

hoechenberger:
> Is the problem specific to using a pool? Is there an alternative to a pool (because in our case we know that we only need one subprocess anyway)?

I just did a quick test with `multiprocessing.Process` instead of `Pool` and it does seem to work indeed!

```python
import multiprocessing
import time


def f():
    print('Executing f()...')
    return None


if __name__ == '__main__':
    process = multiprocessing.Process(target=f)
    process.start()
    time.sleep(1)  # Allow worker to run.

    process.terminate()
    process.join()

    print("Deleting join()'ed Process...")
    del process
```

hoechenberger:
> I just did a quick test with `multiprocessing.Process` instead of `Pool` and it does seem to work indeed!

Also works inside a simple PyQt app.

However, I believe `Process` doesn't support a callback function.

cbrnr:
Interesting, so unless there is some other mechanism that can do the job (i.e. call a function when the process is finished) or I'm missing something obvious (maybe the whole cancelable ICA computation can be implemented differently without `Pool` or even `multiprocessing`) I guess this is really an issue then.

hoechenberger:
> (maybe the whole cancelable ICA computation can be implemented differently without `Pool` or even `multiprocessing`) I guess this is really an issue then.

I suppose we could give Qt threads a shot?

cbrnr:
> I suppose we could give Qt threads a shot?

Start a new Python interpreter and take care of all required (de)serialization in a new thread? Sounds wild. The main issue is that Python already runs in the main process which controls the GUI and does computations. I sometimes think that writing the Qt GUI in C++ and embedding a Python interpreter would have been much easier than writing everything in PyQt.

hoechenberger:
> Start a new Python interpreter and take care of all required (de)serialization in a new thread? Sounds wild.

Isn't that exactly what `multiprocessing` does? Are you saying we'd have to do this manually if using a Qt thread, even when we're doing so through PyQt?

hoechenberger:
Also, would maybe a simple ordinary Python thread suffice? Then we could use `multiprocessing.pool.ThreadPool`.

cbrnr:
> Isn't that exactly what multiprocessing does? Are you saying we'd have to do this manually if using a Qt thread, even when we're doing so through PyQt?

Yes, that's what `multiprocessing` does behind the scenes, and I wouldn't want to do that manually.

> Also, would maybe a simple ordinary Python thread suffice? Then we could use `multiprocessing.pool.ThreadPool`.

I don't think so, we'd run into the GIL which means the GUI would be blocked. That's why I thought a subprocess is the only option (but I might be wrong of course).

hoechenberger:
> I don't think so, we'd run into the GIL which means the GUI would be blocked. That's why I thought a subprocess is the only option (but I might be wrong of course).

I would assume that, since numpy releases the GIL during computations, the GIL shouldn't get in our way at least for ICA computation. But then again, I might be wrong too ;)

hoechenberger:
Ok maybe let's ask on SO first regarding the multiprocessing issue with pythonw.

hoechenberger:
@cbrnr 
I thought we could potentially use [`concurrent.futures.ProcessPoolExecutor`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor), which supports callbacks. However, I discovered it doesn't allow to kill a process once it has been started.

I then came across [Pebble](https://pythonhosted.org/Pebble/), which adds this feature ;) (among others, of course). In any case, I now got the MWE I posted previously working with `python` and `pythonw` now on my Mac!

```python
import pebble
import time


def f():
    print('Executing f()...')
    return None


def callback(x):
    print('Callback invoked')
    return None


if __name__ == '__main__':
    pool = pebble.ProcessPool(max_workers=1)
    process = pool.schedule(f)
    process.add_done_callback(callback)
    pool.close()
    time.sleep(1)  # Allow worker to run.

    process.cancel()
    pool.join()

    print("Deleting join()'ed Pool...")
    del pool
```
Pebble has no external dependencies, and it provides a nice API. I believe it's worth a short trying to see if it solves the `pythonw` issue in MNELAB, WDYT?

cbrnr:
Can you make the required changes to use pebble in this PR?

hoechenberger:
@cbnr it’s all done already!!

hoechenberger:
(At least for ICA)

mnelab/mainwindow.py
```diff
@@ -37,6 +38,7 @@
 __version__ = "0.6.0.dev0"
 
 MAX_RECENT = 6  # maximum number of recent files
+mp.set_start_method("spawn", force=True)  # required for Linux/macOS
```
cbrnr: Could you move this line into `__main__.py` (line 84)? I think this doesn't have to be part of `mainwindow.py`.

cbrnr:
Ah - I don't see commits in the notification view, that's why I missed it. This is great! I wonder why it doesn't work with `multiprocessing` (which is part of the standard lib). It'd be interesting to know why this is the case (and if this could be a bug in `multiprocessing`).

cbrnr:
And should we define a minimum required version of Pebble (e.g. the current version)? I've done that for all other required packages.

cbrnr:
Lastly, I could you add an entry to CHANGELOG.md?

hoechenberger:
> Ah - I don't see commits in the notification view, that's why I missed it. This is great! I wonder why it doesn't work with `multiprocessing` (which is part of the standard lib). It'd be interesting to know why this is the case (and if this could be a bug in `multiprocessing`).

More interestingly, we found it _does_ work with `multiprocessing` processes, just not with their process _pools_. Internally, `Pebble` uses `concurrent.futures`, which in turn uses `multiprocessing`. So basically the only real change we're making when moving to `Pebble` is we're replacing the "pool" (wrapper) managing multiple `multiprocessing` processes.

mnelab/mainwindow.py
hoechenberger: Done in 7587c8b

hoechenberger:
> And should we define a minimum required version of Pebble (e.g. the current version)? I've done that for all other required packages.

I don't know which would be the minimum, so if you're fine with just picking the current release, I can certainly do that!

hoechenberger:
@cbrnr I believe I've addressed all of your comments

hoechenberger:
@cbrnr I didn't check, is ICA the only place where we're currently using multiprocessing pools?

cbrnr:
Yes, ICA is the only place. Thanks @hoechenberger - this is a great change and I learned a lot!

hoechenberger:
> this is a great change and I learned a lot!

Same here!! Let's hope this issue can now finally be considered "resolved" :)

cbrnr:
Yep. Although it's still a workaround. So if Anaconda ever decides to get rid of `pythonw` and distributes a `python` executable which can correctly access the macOS window manager (like the official Python build), we should revisit this workaround because it might not be necessary anymore then.

hoechenberger:
Absolutely!

hoechenberger:
@cbrnr Any plans to draft a new release anytime soon?

cbrnr:
Yes, I'll try to do it sometime next week.

cbrnr:
@hoechenberger I was wondering if we could make this work without `pebble` somehow? Since `pebble` only uses stuff from the standard library, this should be possible, right? We don't even need a pool because we are only spawning one additional process.

I'm asking because I now regret a bit that MNELAB now has one additional dependency just because Anaconda on macOS. And every additional dependency creates work (which I just noticed because I had to create a package for pebble for Arch Linux).

hoechenberger:
Yes we should be able to get it to work with only standard library stuff. Can you open an issue for that?

cbrnr:
Sure, #138.

PR #120, opened 2020-01-30
Accomodate changes to LabRecorder nominal srate

cbrnr:
LabRecorder was modified so that the nominal sampling rate is now a string of a double rather than a string of an int. This is actually correct because the type in the lsl::info structure is in fact a double. This means that XDF files recorded with later versions of LabRecorder (unfortunately I couldn't track down the actual commit that made this change) will throw an error at line 232. This small change should prevent this and it should remain back compatible.

### DIFF
--- a/mnelab/utils/io.py
+++ b/mnelab/utils/io.py
@@ -229,7 +229,7 @@ def parse_chunks(chunks):
                                 hostname=chunk.get("hostname"),  # optional
                                 channel_count=int(chunk["channel_count"]),
                                 channel_format=chunk["channel_format"],
-                                nominal_srate=int(chunk["nominal_srate"])))
+                                nominal_srate=int(float(chunk["nominal_srate"]))))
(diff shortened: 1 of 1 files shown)


cbrnr:
I wonder why `nominal_srate` isn't a float in the first place. So instead of changing this to `nominal_srate=int(float(chunk["nominal_srate"]))` we could just write `nominal_srate=float(chunk["nominal_srate"])`. Could you test if this works with one of the files causing the issue? I will test if this change is backward compatible.

cbrnr:
@dmedine I tested loading several XDF files and it still works. Can you please remove the conversion to `int` and just convert to `float` (I can't make the change myself because I don't have push rights to your repo)?

dmedine:
Done. Hope I got it right...

cbrnr:
Thanks @dmedine!

PR #121, opened 2020-02-01
DOC: Add python.app requirement & Pebble min version to readme

hoechenberger:


cbrnr:
Thanks @hoechenberger!

PR #133, opened 2020-02-12
Add option to plot Time/Frequency maps

### LINKED ISSUE #132: Add option to plot Time/Frequency maps
It would be cool to have a function to easily plot Time/Frequency maps.
Some parameters need to be adjustable:
- [x]  Frequency range (start, stop, resolution)
- [x]  Time range (start, stop)
- [x]  baseline correction (start & stop time, mode)
- [x]  permutation clustering (on / off)

stralu:
Fixes #132.

To Do:
- [ ] The permutation cluster test is very, very slow. Therefore, it should be disabled by default.
- [ ] If computing t/f maps with permutation cluster test enabled, we should perform the computation in a separate process (with the `multiprocessing` module) and let the user cancel the calculation. See the Run ICA dialogs for how this is done.
- [ ] Is it possible to add a time resolution setting as well?
- [x] The statusbar in the resulting figure should be disabled (see the raw plot method where this is already done).
- [x] The menu entry "Time/frequency map..." should have three dots to indicate that it has options.
- [ ] The checkboxes for Frequency and Time are not necessary and should be removed.
- [ ] I'd put the baseline modes in a combobox instead of a list.


stralu:
Can you please take a look onto the Oldest pipeline. Here, the matplotlib==2.0.0 package is used together with PySide2. This setup does not work, as PySide2 was introduced to matplotlib in v2.1.0

cbrnr:
OK, then I'll just bump the minimum required matplotlib version to 2.1.0!

cbrnr:
Done. Please rebase, then the failing job should pass.

stralu:
Nice. What do you think about the code?

mnelab/__main__.py
```diff
@@ -27,7 +28,11 @@ def _run():
                 if info:
                     info["CFBundleName"] = app_name
 
-    matplotlib.use("Qt5Agg")
+    if qtpy.PYQT5 or qtpy.PYSIDE2:
+        matplotlib.use("Qt5Agg")
+    elif qtpy.PYQT4 or qtpy.PYSIDE:
+        matplotlib.use("Qt4Agg")
+
```
cbrnr: I think you can remove these changes (and `import qtpy`), because MNELAB doesn't support the old PyQt4/PySide packages.

mnelab/dialogs/plottimefrequencydialog.py
```diff
@@ -0,0 +1,199 @@
...
+        self.setWindowTitle(title)
+        vbox = QVBoxLayout(self)
+        grid = QGridLayout()
+
+        grid.addWidget(QLabel("From"), 0, 2, Qt.AlignCenter)
+        grid.addWidget(QLabel("Resolution"), 0, 3, 1, 2, Qt.AlignCenter)
+        grid.addWidget(QLabel("To"), 0, 5, Qt.AlignCenter)
+
+        # Frequency
+        grid.addWidget(QLabel("Frequecy (in Hz):"), 1, 0)
```
cbrnr: Typo: Frequecy -> Frequency

mnelab/dialogs/plottimefrequencydialog.py
```diff
@@ -0,0 +1,199 @@
...
+#
+# License: BSD (3-clause)
+
+from qtpy.QtWidgets import (QCheckBox, QDialog, QDialogButtonBox, QDoubleSpinBox,
+                            QGridLayout, QLabel, QListWidget, QListWidgetItem, QVBoxLayout)
+from qtpy.QtCore import Qt, Slot
+
+
+class PlotTFDialog(QDialog):
+    def __init__(self, parent, t_range, f_range, modi, title="Time/Frequency settings"):
```
cbrnr: modi -> modes

mnelab/dialogs/plottimefrequencydialog.py
```diff
@@ -0,0 +1,199 @@
...
+
+        self._lfreq = QDoubleSpinBox()
+        self._lfreq.setMinimum(f_range[0])
+        self._lfreq.setValue(f_range[0])
+        self._lfreq.setMaximum(f_range[1])
+        self._lfreq.setDecimals(1)
+        self._lfreq.setSuffix(" Hz")
+        grid.addWidget(self._lfreq, 1, 2)
+
+        self._freq_reso = QDoubleSpinBox()
```
cbrnr: `self._freq_reso` -> `self._freq_res`

mnelab/dialogs/plottimefrequencydialog.py
```diff
@@ -0,0 +1,199 @@
...
+        self._stop_baseline.setMinimum(t_range[0])
+        self._stop_baseline.setValue(0)
+        self._stop_baseline.setMaximum(t_range[1])
+        self._stop_baseline.setDecimals(2)
+        self._stop_baseline.setSingleStep(0.05)
+        self._stop_baseline.setSuffix(" s")
+        grid.addWidget(self._stop_baseline, 3, 4, 1, 2)
+
+        self._baseline_mode = QListWidget()
+        for modus in modi:
```
cbrnr: `for mode in modes:`

mnelab/mainwindow.py
```diff
@@ -8,20 +8,32 @@
...
-
 import mne
+import qtpy
+import matplotlib
+
+if qtpy.PYQT5 or qtpy.PYSIDE2:
+    matplotlib.use("Qt5Agg")
+elif qtpy.PYQT4 or qtpy.PYSIDE:
+    matplotlib.use("Qt4Agg")
+
```
cbrnr: I think this can be removed.

mnelab/mainwindow.py
```diff
@@ -8,20 +8,32 @@
...
 from mne.io.pick import channel_type
+from mne.viz.utils import center_cmap
+from mne.time_frequency import tfr_multitaper
+from mne.stats import permutation_cluster_1samp_test as pcluster_test
 
 from .dialogs import (AnnotationsDialog, CalcDialog, ChannelPropertiesDialog,
                       CropDialog, AppendDialog, EpochDialog, ErrorMessageBox,
                       EventsDialog, FilterDialog, FindEventsDialog, HistoryDialog,
-                      InterpolateBadsDialog, MontageDialog, PickChannelsDialog,
+                      InterpolateBadsDialog, MontageDialog, PickChannelsDialog, PlotTFDialog,
```
cbrnr: Please make sure not to exceed 79 characters per line.

cbrnr:
I made some quick minor comments because I don't have the time to thoroughly review your code right now. Will do later!

mnelab/__main__.py
```diff
@@ -6,6 +6,7 @@
 import os
 import multiprocessing as mp
 import matplotlib
+import qtpy
```
cbrnr: This import is not needed anymore.

mnelab/__main__.py
```diff
@@ -28,6 +29,7 @@ def _run():
                     info["CFBundleName"] = app_name
 
     matplotlib.use("Qt5Agg")
+
```
cbrnr: This empty line is also not needed.

mnelab/dialogs/plottimefrequencydialog.py
```diff
@@ -0,0 +1,200 @@
...
+# License: BSD (3-clause)
+
+from qtpy.QtWidgets import (QCheckBox, QDialog, QDialogButtonBox,
+                            QDoubleSpinBox,QGridLayout, QLabel, QListWidget,
+                            QListWidgetItem, QVBoxLayout)
+from qtpy.QtCore import Qt, Slot
+
+
+class PlotTFDialog(QDialog):
+    def __init__(self, parent, t_range, f_range, modes, title="Time/Frequency settings"):
```
cbrnr: Can you try to use the PEP8-recommended character limit of 79 everywhere?

mnelab/dialogs/plottimefrequencydialog.py
```diff
@@ -0,0 +1,201 @@
...
+    @Slot()
+    def toggle_baseline(self):
+        if self.baseline_checkbox.isChecked():
+            self._baseline_mode.setEnabled(True)
+            self._start_baseline.setEnabled(True)
+            self._stop_baseline.setEnabled(True)
+        else:
+            self._baseline_mode.setEnabled(False)
+            self._start_baseline.setEnabled(False)
+            self._stop_baseline.setEnabled(False)
```
cbrnr: Missing empty last line.

mnelab/mainwindow.py
```diff
@@ -8,29 +8,35 @@
 from os.path import isfile, isdir
 from functools import partial
 import numpy as np
-
 import mne
+import qtpy
+import matplotlib
+matplotlib.use("Qt5Agg")
```
cbrnr: Did you check if these three lines are really necessary? It seems like they aren't.

mnelab/mainwindow.py
```diff
@@ -8,29 +8,35 @@
...
+from mne.stats import permutation_cluster_1samp_test as pcluster_test
 
 from .dialogs import (AnnotationsDialog, CalcDialog, ChannelPropertiesDialog,
                       CropDialog, AppendDialog, EpochDialog, ErrorMessageBox,
-                      EventsDialog, FilterDialog, FindEventsDialog, HistoryDialog,
-                      InterpolateBadsDialog, MontageDialog, PickChannelsDialog,
-                      ReferenceDialog, RunICADialog, XDFStreamsDialog,
-                      MetaInfoDialog)
+                      EventsDialog, FilterDialog, FindEventsDialog,
+                      HistoryDialog,InterpolateBadsDialog, MontageDialog,
```
cbrnr: Missing space after the comma

mnelab/mainwindow.py
stralu: qtpy is not necessary, but the backend of matplotlib needs to be set

cbrnr:
Alright, I tested this with some example files, and I'd suggest the following changes:

1. The permutation cluster test is very, very slow. Therefore, it should be disabled by default.
2. If computing t/f maps with permutation cluster test enabled, we should perform the computation in a separate process and let the user cancel the calculation. See the Run ICA dialogs for how this is done.
3. Is it possible to add a time resolution setting as well?
4. The statusbar in the resulting figure should be disabled (see the raw plot method where this is already done).
5. The menu entry "Time/frequency map..." should have three dots to indicate that it has options.
6. The checkboxes for Frequency and Time are not necessary and should be removed.
7. I'd put the baseline modes in a combobox instead of a list.

cbrnr:
@stralu I had to remove all code using `pebble` because we've switched back to `multiprocessing`. However, the code is not working yet - do you have time to take a look?

stralu:
Yes of course. But since I am currently fairly busy, it may take a few days.

cbrnr:
No problem, there's no hurry at all! Thanks for taking care of this!

cbrnr:
Continued in #164.

PR #135, opened 2020-02-16
Allow to concatenate raws and epochs

### LINKED ISSUE #131: Append/concatenate epochs
The current append feature only works with continuous raw data. We should support appending epochs via `mne.concatenate_epochs` (and maybe switching from `raw.append` to `mne.concatenate_raws` will make things simpler).

stralu:
Fixes #131.

### DIFF
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -366,7 +366,9 @@ class MainWindow(QMainWindow):
             self.actions["crop"].setEnabled(
                 enabled and self.model.current["dtype"] == "raw")
             append = bool(self.model.get_compatibles())
-            self.actions["append_data"].setEnabled(enabled and append)
+            self.actions["append_data"].setEnabled(enabled and append and
+               (self.model.current["dtype"] == "raw" or
+                self.model.current["dtype"] == "epochs"))
             self.actions["meta_info"].setEnabled(
                 enabled and
                 self.model.current["ftype"] == "Extensible Data Format")
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -367,8 +367,7 @@ class MainWindow(QMainWindow):
                 enabled and self.model.current["dtype"] == "raw")
             append = bool(self.model.get_compatibles())
             self.actions["append_data"].setEnabled(enabled and append and
-               (self.model.current["dtype"] == "raw" or
-                self.model.current["dtype"] == "epochs"))
+               (self.model.current["dtype"] in ("raw", "epochs")))
             self.actions["meta_info"].setEnabled(
                 enabled and
                 self.model.current["ftype"] == "Extensible Data Format")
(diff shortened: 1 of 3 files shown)


mnelab/mainwindow.py
```diff
@@ -366,7 +366,9 @@ def data_changed(self):
             self.actions["crop"].setEnabled(
                 enabled and self.model.current["dtype"] == "raw")
             append = bool(self.model.get_compatibles())
-            self.actions["append_data"].setEnabled(enabled and append)
+            self.actions["append_data"].setEnabled(enabled and append and
+               (self.model.current["dtype"] == "raw" or
+                self.model.current["dtype"] == "epochs"))
```
cbrnr: `self.model.current["dtype"] in ("raw", "epochs")`

mnelab/model.py
```diff
@@ -575,6 +575,8 @@ def get_compatibles(self):
         for idx, d in enumerate(self.data):
             if idx == self.index:  # skip current data set
                 continue
+            if not (d["dtype"] == "raw" or d["dtype"] == "epochs"):
```
cbrnr: `if d["dtype"] not in ("raw", "epochs"):`

mnelab/model.py
```diff
@@ -583,7 +585,8 @@ def get_compatibles(self):
                 continue
             if d["data"].info["bads"] != data.info["bads"]:
                 continue
-            if not all(d["data"]._cals == data._cals):
+            if d["dtype"] == "raw" and \
+                not all(d["data"]._cals == data._cals):
```
cbrnr: `if d["dtype"] == "raw" and any(d["data"]._cals != data._cals):` (also, please try to avoid breaking lines with `\`)

mnelab/model.py
```diff
@@ -598,11 +601,20 @@ def get_compatibles(self):
     @data_changed
     def append_data(self, names):
         """Append the given raw data sets."""
+        files = [self.current["data"]]
         for d in self.data:
             if d["name"] in names and d["data"] is not None:
```
cbrnr: I was wondering when `d["data"]` can be `None`...

cbrnr:
When concatenating epochs, we need to check if all epochs have the same shape (i.e. channels and length), otherwise we can't concatenate. Could you add this to `get_compatibles`?

mnelab/model.py
stralu: Oops, this comes from an older version.

stralu:
We already have the channel check. I added a check for the epoch length (tmin, tmax) and baeline (tmin, tmax).

cbrnr:
Nice! Can you please add an entry to `CHANGELOG.md`? Afterwards, I'll merge.

cbrnr:
Maybe you can just add this PR number to the existing entry under "Added".

cbrnr:
Thanks @stralu!

PR #136, opened 2020-02-17
Improve runicadialog logic

stralu:
Simplified nested containers for the following dialogs:
- append
- crop
- filter
- find events
- reference
- run ica

### DIFF
--- a/mnelab/dialogs/cropdialog.py
+++ b/mnelab/dialogs/cropdialog.py
@@ -2,47 +2,45 @@
 #
 # License: BSD (3-clause)
 
-from qtpy.QtWidgets import (QCheckBox, QDialog, QDialogButtonBox,
-                            QDoubleSpinBox, QFormLayout, QSizePolicy)
-
-from qtpy.QtCore import Qt, Slot
+from qtpy.QtWidgets import (QDialog, QVBoxLayout, QGridLayout, QDoubleSpinBox,
+                            QDialogButtonBox, QCheckBox)
+from qtpy.QtCore import Slot
 
 
 class CropDialog(QDialog):
     def __init__(self, parent, start, stop):
         super().__init__(parent)
         self.setWindowTitle("Crop data")
-
-        form = QFormLayout(self)
-        form.setAlignment(Qt.AlignCenter)
-        label_size = QSizePolicy.Expanding
+        vbox = QVBoxLayout(self)
+        grid = QGridLayout()
         self.start_checkbox = QCheckBox("Start time:")
-        self.start_checkbox.setSizePolicy(label_size, label_size)
         self.start_checkbox.setChecked(True)
         self.start_checkbox.stateChanged.connect(self.toggle_start)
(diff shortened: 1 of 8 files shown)


cbrnr:
It looks like the vertical alignment changes when using `QFormLayout`. The widgets in the right column are too close to the bottom of each row, see the following example (notice that also the horizontal alignment is now right-aligned):

qformlayout

Here's what this dialog looked before (better vertical alignment, better horizontal alignment):

old

Is there a way to change the alignment? Otherwise, I like the old layouts better.

stralu:
Ah, I see. This appearance comes from non-expanding checkboxes. Thus, these two lines of code should help us out:
`label_size = QSizePolicy.Expanding`
`self.start_checkbox.setSizePolicy(label_size, label_size)`

mnelab/dialogs/cropdialog.py
```diff
@@ -14,7 +14,9 @@ def __init__(self, parent, start, stop):
         self.setWindowTitle("Crop data")
 
         form = QFormLayout(self)
+        label_size = QSizePolicy.Expanding
```
cbrnr: I think you forgot to import `QSizePolicy`.

stralu: Yeah, I didn't check this line in.

stralu: What do you think about this solution. Should we really prefer it over a QGridLayout?

cbrnr: Not convinced yet. The vertical alignment seems to be OK, but the horizontal isn't right for items containing checkboxes. On macOS, the left column is by default right-aligned, this looks good if there are only labels, but with checkboxes not so much.

Screen Shot 2020-02-17 at 17 03 34

stralu: Interesting, it perfectly fits on Ubuntu.
image

cbrnr: Yes, it's because `QFormLayout` tries to adhere to each platform's interface guidelines, and on macOS the guidelines say that the labels must be right-aligned (https://doc.qt.io/qt-5/qformlayout.html). So I think we should use `QFormLayout` only for dialogs where the labels are just... labels.

mnelab/dialogs/cropdialog.py
```diff
@@ -14,6 +14,7 @@ def __init__(self, parent, start, stop):
         self.setWindowTitle("Crop data")
 
         form = QFormLayout(self)
+        form.setAlignment(Qt.AlignCenter)
```
stralu: Can you try this?

cbrnr: I think we shouldn't mess around with the layouts too much. Let's only use `QFormLayout` for the simplest dialogs. BTW, KDE also aligns the labels to the right.

cbrnr:
Thanks @stralu!

Issue #139, opened 2020-02-28
Use custom startup script for Anaconda on macOS


cbrnr:
#119 introduced a workaround to run the GUI properly when using Anaconda on macOS (basically, it spawns a new process and starts the GUI with `pythonw` if it was started with `python`).

I think we should probably revert this change, and instead create a custom startup script for Anaconda. People don't have to start MNELAB with `python -m mnelab` &ndash; we could create a startup script instead. This would automatically solve the multiprocessing problem due to differing Python versions in Anaconda's `python` and `pythonw` executables (#138). What do you think @hoechenberger?

cbrnr:
Maybe this won't be easy. I forgot that `setup.py` already contains

```python
    entry_points={
        'gui_scripts': [
            'mnelab=mnelab.__main__:main',
        ],
```

This automatically creates an executable `mnelab` script in `/usr/local/bin` (usually) after `pip install mnelab`, but this script will contain the Python executable used to call `pip` (usually `python -m pip install mnelab`). I've read somewhere that `pythonw -m pip install mnelab` doesn't work either, so I guess we're stuck with the current workaround to spawn a new process. Any additional ideas how this workaround could be removed/outsourced somewhere where normal Python users are not confronted with it? I really liked the previous simple way to start the app!

cbrnr:
The Anaconda workaround is gone and I don't want to explicitly support Anaconda. It works with the official Python distribution (including Homebrew).

PR #140, opened 2020-03-05
Remove pebble again

### LINKED ISSUE #138: Replace pebble with standard library modules
Currently we require `pebble` because `multiprocessing.Pool` did not work with the workaround introduced for Anaconda on macOS (#119). In theory, it should be possible to make do without `pebble`, because this package is just a wrapper for some standard library modules.

cbrnr:
Fixes #138. Needs testing on:

- [x] macOS (conda)
- [x] macOS (regular Python)
- [x] Windows
- [x] Linux


cbrnr:
That's weird. Now I get

```
/Users/clemens/anaconda/envs/python38/bin/pythonw: line 3:  5248 Illegal instruction: 4  /Users/clemens/anaconda/envs/python38/python.app/Contents/MacOS/python "$@"
```

even on `master`. I'll try to figure out what the problem is.

cbrnr:
Maybe something's wrong with my `conda` setup because I can't even run older MNELAB versions such as v0.5.2. @hoechenberger do you have time to test this PR on your machine?

cbrnr:
I actually get this segfault after `import matplotlib`... I'm pretty sure this must be my particular setup (I'm running the latest macOS on an unsupported MacBook).

cbrnr:
Yeah, it's probably because my CPU doesn't have AVX or something, but conda packages are compiled with this feature on. I will test on a different machine, but now I'm pretty sure it will work. I also assume that conda will work when I roll back to the last supported macOS on this older machine - does anyone know how and why this works (the binary conda package is the same, does the OS perform some conversions when it doesn't know a certain instruction set)?

hoechenberger:
@cbrnr I can test this non my computer in a couple hours

hoechenberger:
I don’t understand how this is working now when the exact same thing broke conda a couple of weeks ago?

cbrnr:
@hoechenberger I just tested it on my other MacBook, and everything works perfectly fine! The reason why it works now is because Anaconda has now updated `python.app` for all Python versions so that `python` and `pythonw` are now the same versions. This was the reason for the segfaults.

cbrnr:
Please go ahead and test it on your machine as well - once I have your OK I will merge.

hoechenberger:
Oh wow they did??? I didn’t realize!!! Awesome!

hoechenberger:
Ok. Since I’m typically using conda-forge I’ll test this separately and open an issue on the conda-forge issue tracker if need be.

cbrnr:
Alright, but can you please report back here if it works? I don't want to merge before you have tested this (even with conda-forge, which I have also already tested on my laptop and it works).

hoechenberger:
Will get back to you until tonight.

cbrnr:
Then the next step will be #139, in which I'd like to create an `mnelab` startup script which contains the logic/special case for conda on macOS.

hoechenberger:
Yes, got it. :)

hoechenberger:
@cbrnr I tested with `conda-forge` and it does work with their latest Python 3.6 and 3.8 builds, but crashes with Python 3.7. `python` and `pythonw` version numbers do not match for either Python version.

cbrnr:
OK, so all Python versions (3.6, 3.7, 3.8) in the default channel (or whatever it is called) work (and `python` and `pythonw` versions match).

In conda-forge only 3.6 and 3.8 work, and `python` and `pythonw` versions don't match (not even for 3.6 and 3.8). This raises two questions:

1. What's the default version of Python when you install the conda-forge package?
2. Should we open an issue for the python and python.app packages on conda-forge because their versions don't match (but should)?

Did I summarize this correctly?

hoechenberger:
> Did I summarize this correctly?

Yes, absolutely.

> 1. What's the default version of Python when you install the conda-forge package?

I believe it tries to install the latest version of any package that satisfies the dependency tree. It would resolve to Python 3.8.2 for now.

> 2. Should we open an issue for the python and python.app packages on conda-forge because their versions don't match (but should)?

Yes, definitely.

cbrnr:
> I believe it tries to install the latest version of any package that satisfies the dependency tree. It would resolve to Python 3.8.2 for now.

OK, so this means `conda create -n mnelab mnelab` should install Python 3.8.2, which does work with reverting to `multiprocessing`. Then we should good to go?

I'll open an issue regarding the version mismatch.

hoechenberger:
> I'll open an issue regarding the version mismatch.

I already did: 
https://github.com/conda-forge/python.app-feedstock/issues/8



> OK, so this means `conda create -n mnelab mnelab` should install Python 3.8.2, which does work with reverting to `multiprocessing`. Then we should be good to go?

To be precise `conda create -c conda-forge -n mnelab mnelab`; otherwise +1 on what you said.

cbrnr:
Saw the issue, thanks!

Yes, I forgot the channel 🤦‍♂. If I ever have to use conda, I think I'll switch to conda-forge because the defaults channels stopped working for my MacBook Pro Mid-2010: https://github.com/conda/conda/issues/9678. I assume they introduced this by accident so maybe they'll revert it. Otherwise yet another reason for me to not use conda at all.

Issue #34, opened 2019-07-03
Average epochs


mmagnuski:
Summary of feature requests from this discussion:
- [x] epoching (fixed in #45)
- [x] epoch selection (by condition/event)
- [x] epochs concatenation (will be fixed in #131)
- [x] epochs averaging (evoked)
- [x] ERP plotting - this should work once epochs averaging is implemented

mmagnuski:
sorry, accidental posting. :)
I wanted to ask about epoching - it seems not to be available in mnelab.
More broadly it would be good to have a list o mnelab features somewhere in the docs as it is not easy to tell what can be done in mnelab just by looking into main repo the documentation.

To use mnelab in class the following features would be necessary (at least from my perspective):
* epoching
* epoch selection (by condition/event)
* epochs averaging (evoked)
* ERP plotting
* creating new data by applying ICA 

are these features within future scope of mnelab? If so I would like to help at some point but first I'd have to teach myself some pyqt. :)

mmagnuski:
and it would be super useful to display code responsible for certain GUI operations (the equivalence of `eeghist` in eeglab).

cbrnr:
I don't have a detailed roadmap for MNELAB, but epoching is definitely at the top of the list. This will require some restructuring (or at least thinking), so I can't say when it will be done. Applying ICA should be easier since computing ICA is already possible, and all that's missing is a simple command.

Something like `eeghist` already exists in the code, I just need to add a way to access the command history.

cbrnr:
And regarding the documentation: there is no documentation yet, this would probably be the most important thing to start adding...

vferat:
Some of these features are implemented in the fork we work on with @TyWR. 

Some of them such as epoching and averaging are working fine, but the structure of the code may not be optimal for long term dev. For others features, some of them are custom plot that may not suit the MNE community. But there is several functions such as ICA and filtering which are pure MNE code that could be merged to the main repo.

In addition, we can help to developt new functions that are in neither of the repos

cbrnr:
@vferat PRs welcome - maybe start with applying ICA to raw data? I will also work on some stuff, we should probably coordinate.

mmagnuski:
@cbrnr 
Oh, I now see that the command history is printed in the console, great!

hoechenberger:
Two additional features that are essential for my workflow:
- [concatenate epochs](mne.concatenate_epochs)
[...]

PR #145, opened 2020-05-06
MRG: Add NIRS support

rob-luke:
This pull request provides NIRS support for MNELAB

### Rough list of features to test before requesting merge

- [x] Open a file without error
- [x] View waveform of raw data
- [x] View optode position information (montage)
- [x] Convert data to optical density then haemoglobin
- [x] Deactivate menu items that are not appropriate for data type

~~Apply filtering and noise reduction~~ Removed in PR rescope. 
~~View epoched data~~ Removed in favour of #34

### Status

Just started on this project. I will keep working in this branch and will squash commits before requesting a review.


cbrnr:
I'm fine with bumping the required MNE version to 0.20. I'll do that in a separate PR, then you can just rebase this PR.

cbrnr:
Alright, you can go ahead and rebase.

cbrnr:
I think at some point we will need a mechanism to determine file types not based on their extensions. Currently I don't think we already have duplicate extensions, but `.hdr` sounds like other file types could be using it as well. I'll create an issue.

rob-luke:
> Alright, you can go ahead and rebase.

Thanks @cbrnr 

I am a bit stuck trying to get the montage button to work with the nirs data. The montage is automatically loaded when you initially load the data, so I want the button to be enabled (not greyed out). If I force the montage variable [here](https://github.com/cbrnr/mnelab/blob/master/mnelab/mainwindow.py#L349) to be true then I can click the button and the plot looks correct. But I can't find an elegant way to set the montage variable to true when loading nirs data. Can you suggest an approach?

Thanks for the help, GUI programming is new to me.

cbrnr:
Can you set the montage using `model.set_montage`? This does two things: (1) it calls `self.current["data"].set_montage`, and (2) it sets `self.current["montage"]` to the `montage`. MNELAB needs to keep track of applied montages because MNE doesn't do it.

mnelab/model.py
```diff
@@ -145,6 +147,9 @@ def load(self, fname, *args, **kwargs):
                                      ftype=ftype, fsize=fsize, data=data,
                                      dtype=dtype))
 
+        if ext == ".hdr":
+            self.current["montage"] = True
```
rob-luke, line 151: Thanks for the pointer @cbrnr. This works, but I still feel a cleaner solution may exist. I will revisit this as I get a better understanding of MNELAB.

cbrnr: Yes, I was thinking that you could explicitly call `self.model.set_montage` with the correct `montage` object. The problem here seems to be that such a `montage` object doesn't exists as the channel locations are directly set in the loading function. Do you think it is possible to extract the montage part in `read_raw_nirx` into a separate step?

cbrnr: To clarify, I mean calling `set_montage` inside `_load_nirx`.

rob-luke: I think I will have to do some more reading of the MNE manuals. I think I have misunderstood what is meant by the term montage. The location of sensors (optodes) is stored in `info` in the `locs` array. I was assuming this is what was meant by a montage. I will look in to the MNE docs.

However, the code above enables the button in the MNELAB main window. If I click the button it displays the sensors in the correct position on a 2d head. So this is the outcome I desired. Now I just need to make the code correct :)

cbrnr: I never seem to remember where MNE stores channel locations. According to the docs, it should be in `info["chs"]`, which is a list of dicts for each channel. Each dict has a `"loc"` key, which contains the coordinates of that channel. However, there's also `info["dig"]`, which is also a list of dicts for each channel. Each dict has an `"r"` key containing channel coordinates. Not sure what the difference between these two entries is, but I think the first one is used for montages.

When you do `raw.set_montage(...)`, MNE populates the `info["chs"]` list of dicts (specifically the `"loc"` values). But there's no way to determine the montage given the locations - which is why I'm saving the montage explicitly (partly also because MNELAB displays the montage name).

cbrnr: According to [this tutorial](https://mne.tools/dev/auto_tutorials/intro/plot_40_sensor_locations.html), both `chs` and `dig` entries are populated. Maybe MNELAB should determine whether a montage is set (or in other words, if channel locations are available) by looking at these two keys. Then it would be unnecessary to store that extra information in `self.current["montage"]`, because this information would be available in `self.current["data"].info`...

cbrnr: I've implemented this in #147. If you rebase, you can remove setting the montage explicitly, this should now be picked up automatically.

rob-luke: Thanks. I appreciate your help here.

I will move on to adding the two required conversion steps next.

rob-luke:
I have removed "View epoched data" from the initial list of tasks as this is already in progress at #34.

I have also removed filtering and noise reduction as I think that is better suited to a follow up PR. I am keeping this PR to basic usage of opening NIRS files, visualising them to check data quality, and exporting as fif.

rob-luke:
This now has the basic features I was after. I will use it for a few days before requesting a merge. But it could do with a review whenever you have time @cbrnr.

cbrnr:
@rob-luke I've refactored the readers code. It should now be a bit easier to add new file formats. NIRX (.hdr) is already activated in master. Could you please rebase?

rob-luke:
> Could you please rebase?

Done. Thanks for your help, its greatly appreciated.

Given your improvements above, this PR is about enabling the required conversion steps for NIRS processing. This involves converting the raw data to optical density, and then haemoglobin concentration.

I have currently placed these options in their own "section" (not sure what correct GUI term to use here) within the edit dropdown (see below). Is this a good place for them? Im happy to put them wherever you think is best.

image

Also, I would like to add a few more NIRS options, specifically the [temporal derivative distribution repair (TDDR) ](https://mne.tools/dev/generated/mne.preprocessing.nirs.temporal_derivative_distribution_repair.html). Could I add these to the same "section" in the edit file? If you think it will get too messy then I could create a NIRS specific sub menu (similar to your current export data in file).

cbrnr:
Where to put the actions is a really good question. I like the idea of having a NIRS submenu in the Tools menu. Alternatively, we could also create a top-level NIRS menu entry if there are going to be a lot of NIRS-specific things. The current menu structure is WIP, I'm pretty sure we will have to restructure it as the project progresses (e.g. create a separate Channels menu, maybe an ICA menu, ...).

For now, it'd probably be best to use a NIRS submenu within Tools. We can always create a NIRS top-level menu later on.

Also, I think it makes sense to create a new data set when converting to OD and Hb/HbO2. You can use `self.auto_duplicate()` (see e.g. `filter_data` on how it's done).

Finally, please make sure that lines are <= 79 characters.

rob-luke:
Thanks for the feedback @cbrnr. I have made the changes you suggested. I will work with this for a week and see how it goes before requesting a merge.

mnelab/mainwindow.py
```diff
@@ -222,6 +222,17 @@ def __init__(self, model):
         self.actions["events_from_annotations"] = tools_menu.addAction(
             "Create events from annotations", self.events_from_annotations
         )
+        tools_menu.addSeparator()
+        self.export_menu = tools_menu.addMenu("NIRS")
```
cbrnr, line 226: This should be called `self.nirs_menu`.

mnelab/mainwindow.py
```diff
@@ -222,6 +222,17 @@ def __init__(self, model):
...
+        self.actions["convert_od"] = self.export_menu.addAction(
+                "Convert &to optical density", self.convert_od)
```
cbrnr, line 228: We need to have unique mnemonics, i.e. &t is used by all three menu entries here. I'd put them in front of some meaningful letter, e.g. &density here.

mnelab/mainwindow.py
```diff
@@ -222,6 +222,17 @@ def __init__(self, model):
...
+        self.export_menu.addSeparator()
+        self.actions["apply_tddr"] = self.export_menu.addAction(
+                "Apply &TDDR", self.apply_tddr)
+        self.export_menu.addSeparator()
+        self.actions["convert_bl"] = self.export_menu.addAction(
+                "Convert &to haemoglobin", self.convert_bl)
```
cbrnr, line 234: &haemoglobin

mnelab/model.py
```diff
@@ -443,6 +443,31 @@ def epoch_data(self, events, tmin, tmax, baseline):
...
+        self.current["data"] = mne.preprocessing.nirs.optical_density(
+            self.current["data"])
+        self.current["name"] += f" (OD)"
+        self.history.append(
+            f'data = mne.preprocessing.nirs.optical_density(data)')
+
+    @data_changed
+    def apply_tddr(self):
+        self.current["data"] = mne.preprocessing.nirs.\
+            temporal_derivative_distribution_repair(self.current["data"])
```
cbrnr, line 457: Maybe import this function with a shorter name? E.g. `tddr`?

rob-luke: If I make it `tddr` this will still cross two lines. Plus there will be the added line for the import. So this is less lines in total and is more explicit. But happy to add the import, do you prefer all imports up the top, or should I nest the import in `apply_tddr`?

mnelab/mainwindow.py
```diff
@@ -222,6 +222,17 @@ def __init__(self, model):
         self.actions["events_from_annotations"] = tools_menu.addAction(
             "Create events from annotations", self.events_from_annotations
         )
+        tools_menu.addSeparator()
+        self.nirs_menu = tools_menu.addMenu("NIRS")
```
cbrnr, line 226: We don't need an attribute for this menu because we never access it except here in `__init__`. So just use `nirs_menu` instead of `self.nirs_menu`.

mnelab/model.py
cbrnr: If you import the function as `from mne.preprocessing.nirs import temporal_derivative_distribution_repair as tddr` up top this will fit into one line.

I'd not do this for the history entry though, because fiddling with imports in the history will complicate things. Therefore, the history entry is fine as it is.

However, would it make sense to rename this function to `tddr` directly in MNE? We often use abbreviations, and given that this function is in the `nirs` subpackage, it should be clear to NIRS people what this function does (otherwise there's still the docstring).

mnelab/model.py
```diff
@@ -443,6 +443,31 @@ def epoch_data(self, events, tmin, tmax, baseline):
...
+        self.current["name"] += f" (TDDR)"
+        self.history.append(
+            f'data = mne.preprocessing.nirs.'
+            f'temporal_derivative_distribution_repair(data)')
```
cbrnr, line 461: For consistency, I'd prefer double-quotes for strings wherever possible. Same for the history f-string in `convert_beer_lambert` below.

mnelab/model.py
```diff
@@ -443,6 +443,31 @@ def epoch_data(self, events, tmin, tmax, baseline):
...
+
+    @data_changed
+    def convert_beer_lambert(self):
+        self.current["data"] = mne.preprocessing.nirs.beer_lambert_law(
```
cbrnr, line 465: Same comment here, do you think it would make sense to rename this function to `beer_lambert` directly in MNE?

mnelab/model.py
rob-luke: Great I will make the changes as you suggest.

Regarding changing the name in MNE, I orginally submitted the code as a PR with `tddr` naming, but was asked to change it to what it is now. So I think we just leave it.

cbrnr: OK. I'm sure that you could convince others that the abbreviation is a much better name, but that's your call. I'm fine with leaving this for now.

rob-luke:
Should I remove TDDR completely and add back in after next MNE release? Or I should change to the new short name and we just don't merge until after new MNE release?

I think Im the only one using this, so I'm fine either way.

cbrnr:
I think it would be great to have some NIRS functionality in MNELAB before the next MNE release (which is going to be sometime in September/October). Therefore, I'd remove TDDR for now and we can merge everything else. Can you create a new issue to remind ourselves of including TDDR as soon as it becomes available in MNE? And I'd import `tddr` then.

cbrnr:
LGTM! Can you please add a CHANGELOG entry? Then if you're happy with the PR this is ready to be merged.

cbrnr:
@rob-luke if you're happy with the changes and after you've added a changelog entry I'd like to merge this PR. Could you take a final look?

rob-luke:
Sorry I'll get to this ASAP

rob-luke:
Change log added and I made a final check of the code.

Thanks for all your assistance @cbrnr getting this ready.

rob-luke:
Sorry @cbrnr I don't understand why the tests have failed. Could you please restart them, perhaps its a CI fault.

cbrnr:
All Linux tests fail because there seems to be a problem with the framebuffer. Probably due to an update to either some library or pytest. We can ignore these errors (but I'll try to fix them eventually).

Thanks @rob-luke - looking forward to further NIRS extensions to MNELAB!

rob-luke:
🎉 🎉 🎉 🎉 🎉  Thanks so much @cbrnr. This tool has already been a great benefit in our lab. 

I can't see a citation pointer in the readme, but I will be sure to find another relevant paper of yours to cite for any work resulting from this tool.

cbrnr:
Thanks @rob-luke - currently there is no paper on MNELAB, but I think it might be timely to write a short introduction now that it has a reasonable amount of basic features.

rob-luke:
Well I'll try and be the first to cite it then.

PR #148, opened 2020-05-06
Refactor readers

cbrnr:
This PR refactors the reading functions into a more generic and separate `io` package.

TODO:
- [x] What about history entries?
- [x] Support not only `raw` but also `epoch` and `evoked`
- [x] Additional parameters e.g. for XDF (stream selection)
- [x] Catch errors (and display dialog)


cbrnr:
I'm using `mnelab.io.read_raw` as history entries, because extracting the original `mne.io.read_raw_*` functions makes things so much uglier.

Also, currently only continuous data (aka raw) works. This could be a problem for formats that support both continuous and epoched (and evoked) data (e.g. EEGLAB .set, FIF). I have to think about how to handle such formats. Previously, FIF epochs and FIF evoked worked because I had a try/except block, which just tried to load raw/epoched/evoked data sequentially.

cbrnr:
Maybe the dict entries of supported readers/writers could be 2-tuples, the first entry is the function and the second entry is the function name (as a string).

PR #152, opened 2020-06-03
Automatically detect light/dark mode (on macOS)

### LINKED ISSUE #151: Adapt icons to dark mode
Currently MNELAB icons look good only in light mode (text works in both modes because it adapts). In dark mode, icons should be inverted. I don't know if Qt supports automatic dark/light mode settings though.

Screen Shot 2020-06-01 at 15 04 14

cbrnr:
Fixes #151.

What's missing in this PR is to implement the actual icon set change whenever a palette change is detected (which means I need to add an inverted icon theme).

I think this should also work on Windows 10 and GNOME (Linux), but I haven't researched the required system calls. Maybe Qt even decides to include such a functionality in their API, which would make all of this much simpler.


GuillaumeFavelier:
I tried to find something generic with Qt without any success so far...

BTW, the failures on Azure for Linux seem to be related to an [API change in `PyVirtualDisplay`](https://github.com/The-Compiler/pytest-xvfb/issues/22#issuecomment-636000864) having a snowball effect on `pytest-xvfb`.

cbrnr:
> I tried to find something generic with Qt without any success so far...

It already works (at least on macOS). I wish that Qt supported this in their API though.

> BTW, the failures on Azure for Linux seem to be related to an API change in PyVirtualDisplay having a snowball effect on pytest-xvfb.

Thanks - do you know if the pytest-xvfb devs already know about that?

GuillaumeFavelier:
> do you know if the pytest-xvfb devs already know about that?

[Yes they are](https://github.com/The-Compiler/pytest-xvfb/issues/22#issuecomment-636922245), I believe this should be fixed in a near future.

The-Compiler:
pytest-xvfb v2.0.0 is [released](https://github.com/The-Compiler/pytest-xvfb/releases/tag/v2.0.0) now - thanks everyone for your patience, and let me know if any issues come up with the new release!

GuillaumeFavelier:
Congratulations for the release :)

cbrnr:
Thanks @The-Compiler!

cbrnr:
@The-Compiler I'm still getting errors on Linux - is this still the same error or something else?
https://dev.azure.com/clemensbrunner/clemensbrunner/_build/results?buildId=668&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=6c939d89-0d1a-51f2-8b30-091a7a82e98c

The-Compiler:
@cbrnr That's unrelated to pytest-xvfb - my guess is that you don't have the necessary dependencies needed by Qt 5.15 installed. Make sure you install `libxcb-icccm4 libxcb-image0 libxcb-keysyms1 libxcb-randr0 libxcb-render-util0 libxcb-xinerama0 libxcb-xfixes0` or see the example config in [pytest-qt's docs](https://github.com/pytest-dev/pytest-qt/blob/master/docs/troubleshooting.rst#github-actions) (though that's for GitHub Actions).

If that isn't it, you can run pytest with `--capture=no --no-qt-log` to find out more.

cbrnr:
Thanks, that's very helpful!

cbrnr:
Installing the missing packages fixed it!

PR #164, opened 2020-09-03
Add option to plot Time/Frequency maps

### LINKED ISSUE #132: Add option to plot Time/Frequency maps

abcsds:
Continuing work to fix #132, from PR #133

- [x]  The permutation cluster test is very, very slow. Therefore, it should be disabled by default.
- [x] If computing t/f maps with permutation cluster test enabled, we should perform the computation in a separate process (with the multiprocessing module) and let the user cancel the calculation. See the Run ICA dialogs for how this is done.
- [x] Is it possible to add a time resolution setting as well?
- [x] The statusbar in the resulting figure should be disabled (see the raw plot method where this is already done). **Disabled by default**
- [x] The menu entry "Time/frequency map..." should have three dots to indicate that it has options.
- [x] I'd put the baseline modes in a combobox instead of a list.

Time/Frequency maps are currently functional, but the dialogue is to be improved.


abcsds:
The clustering algorithm is now working in parallel.
I don't see any checkboxes for Frequency and Time. Have they been removed already?
What exactly is the time resolution setting? neither the `tfr_morlet` nor the `tfr_multitaper` methods have such an option.
Could it be the decimals used to select the filter resolution? The dialogue can be modified to increase the decimal resolution. The trade off is having to click many times when changing the integers. What would you suggest, @cbrnr?

cbrnr:
Thanks @abcsds for taking over this PR. I've tried it and the cluster permutation algorithm seems to be working with multiprocessing.

> I don't see any checkboxes for Frequency and Time. Have they been removed already?

Yes.

> What exactly is the time resolution setting? neither the `tfr_morlet` nor the `tfr_multitaper` methods have such an option.

You're right, I think I was referring to the `decim` parameter, which we could include in the dialog (and set it to `1` by default to not perform decimation).

> Could it be the decimals used to select the filter resolution? The dialogue can be modified to increase the decimal resolution. The trade off is having to click many times when changing the integers.

That's a different issue. I wouldn't change the resolution of the spinboxes.

I have some additional comments/questions I would like to discuss.

1. I'd move the "Method" combobox to the bottom of the dialog because this seems like the least important choice for users.
2. How is cropping handled now? In the [ERDS maps example](https://mne.tools/dev/auto_examples/time_frequency/plot_time_frequency_erds.html), I create epochs that are 0.5s longer on both ends, compute the TFRs, and then crop the 0.5s from the beginning and end. How is this handled here? Do users need to create longer epochs by themselves? If so, we'd need a crop option in the dialog (e.g. crop Xs from the beginning and Ys from the end).
3. Regarding the multiprocessed cluster test, is this parallelized over channels or event types? The original intention was that TFR maps from different channels are completely independent and therefore this can be parallelized.

cbrnr:
Also, if you want to pull in the latest master changes could you please rebase instead of merging? That's how I usually do it (even though I squash and merge PRs so technically this doesn't even make a difference), and I find the commit history easier to follow without merge commits.

abcsds:
1. I've moved the "method" combo box to the bottom of the dialogue.
2. Yes, the user must create longer epochs, and the cropping gets done in the time indicated by the variables `_start_time` and `_stop_time`. These are already available to the user through the dialogue, labeled as "[Time (in s):](https://github.com/abcsds/mnelab/blob/tf_maps/mnelab/dialogs/timefrequencydialog.py#L48)"
3. The parallelization is over events. It's how it was programmed in the last PR. Should I change it to per-channel parallelization? I'd gladly leave it like that because I haven't seen any difference between parallelized and non-parallelized code, but I also don't have a big example file to test it with.

Understood. I'll rebase from now on. I just had no preference before.

abcsds:
Hey @stralu! I understand you want to take over this pr. That's great! Unfortunately I think gh [doesn't have a way to give you access to it](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork), but you can fork it and crate a new one.

PR #166, opened 2020-09-07
Annotations from events

### LINKED ISSUE #114: Add option to convert events to annotations
Currently MNELAB can convert annotations to events, but not the other way round.

abcsds:
Add GUI options to create data annotations from events. Fixes #114.


cbrnr:
Thanks @abcsds! This PR gave me a new idea how we could further improve the events/annotations business. Specifically, `mne.events_from_annotations` returns a tuple with the events and a dictionary mapping events IDs to descriptions. That way, no information is lost in the conversion, and we could use this additional mapping when converting back from events to annotations. WDYT?

cbrnr:
I've created a new issue #167 to discuss this idea.

Issue #167, opened 2020-09-15
Preserve annotation descriptions when converting to events


cbrnr:
Currently, we lose annotation descriptions when converting annotations to events. However, `mne.events_from_annotations` returns a tuple with the first element containing the event array and the second element containing the mapping from annotation description to event ID. I suggest that we store this tuple in `model.current["events"]` instead of only the event array. In cases where this mapping is not available, we could still use a 1-tuple and only store the event array.

abcsds:
Saving the event mappings as a tupple on `model.current["events"]` would require a change on every method that uses the current `model.current["events"]`. Wouldn't it be better to create a new variable for example `model.current["event_mappings"]` and use it when converting from events to annotations. This is a simpler fix. We don't really have any other use for the mappings apart from converting back to annotations, do we?

cbrnr:
It's only used in 5 places, so these changes would be doable. But I agree your suggestion is a good alternative and doesn't require any changes so we might as well go for that.

abcsds:
Here is my proposal:
I'm just checking if the variable has been created before changing back to annotations.
To write the history I used a multi-line string. I did this to keep the indentation of the if statement, but I just noticed that you do the check for the variable outside the history, and keep the history one line (I saw it in the `plot_data` function of the main window. I'm changing my PR to that.

PR #170, opened 2020-09-25
Event mappings Fix for #167

abcsds:
Store event mappings when calling `mne.events_from_annotations` to be able to use them latter on. This commit also considers that event_mappings should be a dictionary of `event_id` and event descriptions with `event_id` as an integer key of a dictionary, and the event description a string. contrary to how they are returned by the `mne.events_from_annotations` function. A fix for this has been implemented by swapping keys and values.

The `event_mappings` is originally swapped because it can be used when [creating epochs](https://mne.tools/stable/generated/mne.Epochs.html#mne.Epochs). If we plan to use the event mappings here I can remove the swap, and implement it only when converting back to annotations. Currently we only use the `mne.Epochs` in a [single function](https://github.com/cbrnr/mnelab/blob/cce5d7c9fc8cf15bca7f9c600d63647048265ca4/mnelab/model.py#L446) which only uses a list of integer events as far as I know. From the documentation I understand that this list can also be the event_mappings dictionary, instead of the list of events. Did I get it right? Can we use the `event_mappings` for epoching?

### DIFF
--- a/mnelab/model.py
+++ b/mnelab/model.py
@@ -131,8 +131,7 @@ class Model:
     @data_changed
     def events_from_annotations(self):
         """Convert annotations to events."""
-        events, event_mappings = mne.events_from_annotations(
-                                    self.current["data"])
+        events, event_mappings = mne.events_from_annotations(self.current["data"])
         # Mappings for annots are swapped {str: int} should be {int: str}
         event_mappings = dict((v, k) for k, v in event_mappings.items())
         if events.shape[0] > 0:
@@ -140,7 +139,8 @@ class Model:
             self.current["event_mappings"] = event_mappings
             self.history.append("events, event_mappings = "
                                 "mne.events_from_annotations(data)"
-                                "event_mappings = dict((v, k) for k, v in event_mappings.items())")
+                                "event_mappings = "
+                                "dict((v, k) for k, v in event_mappings.items())")
 
     @data_changed
     def annotations_from_events(self):
@@ -155,11 +155,10 @@ class Model:
         if len(annots) > 0:
             self.current["data"].set_annotations(annots)
             hist = ('annots = mne.annotations_from_events(data,'
-                    'data.info["sfreq"],')
+                    'data.info["sfreq"]')
             if "event_mappings" in self.current:
-                hist += "event_mappings)"
-            else:
-                hist += ')'
+                hist += ", event_mappings"
+            hist += ')'
             self.history.append(hist)
             self.history.append("data = data.set_annotations(annots)")
 
--- a/mnelab/model.py
+++ b/mnelab/model.py
@@ -131,23 +131,38 @@ class Model:
     @data_changed
     def events_from_annotations(self):
         """Convert annotations to events."""
-        events, _ = mne.events_from_annotations(self.current["data"])
+        events, mappings = mne.events_from_annotations(self.current["data"])
         if events.shape[0] > 0:
             self.current["events"] = events
-            self.history.append("events, _ = "
+            self.current["event_mappings"] = mappings
+            self.history.append("events, event_mappings = "
                                 "mne.events_from_annotations(data)")
 
     @data_changed
     def annotations_from_events(self):
         """Convert events to annotations."""
-        annots = mne.annotations_from_events(
-            self.current["events"],
-            self.current["data"].info["sfreq"]
-        )
+        if "event_mappings" in self.current:
+            annots = mne.annotations_from_events(
+                self.current["events"],
+                self.current["data"].info["sfreq"],
+                event_desc=self.current["event_mappings"]
+            )
+        else:
+            annots = mne.annotations_from_events(
+                self.current["events"],
+                self.current["data"].info["sfreq"],
+            )
         if len(annots) > 0:
             self.current["data"].set_annotations(annots)
-            self.history.append("annots = mne.events_from_annotations(data)")
-            self.history.append("data = data.set_annotations(annots)")
+            self.history.append("""
+if event_mappings is None:
+    annots = mne.annotations_from_events(data, data.info["sfreq"])
+else:
+    annots = mne.annotations_from_events(data,
+                                         data.info["sfreq"],
+                                         event_mappings)
+data = data.set_annotations(annots)
+""")
 
     def export_data(self, fname, ffilter):
         """Export raw to file."""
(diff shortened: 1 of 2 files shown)


mnelab/model.py
```diff
@@ -131,22 +131,39 @@ def find_events(self, stim_channel, consecutive=True, initial_event=True,
...
-        events, _ = mne.events_from_annotations(self.current["data"])
+        events, mappings = mne.events_from_annotations(self.current["data"])
+        # Mappings for annots are swapped {str: int} should be {int: str}
+        mappings = dict((v, k) for k, v in mappings.items())
         if events.shape[0] > 0:
             self.current["events"] = events
-            self.history.append("events, _ = "
+            self.current["event_mappings"] = mappings
+            self.history.append("events, event_mappings = "
                                 "mne.events_from_annotations(data)")
```
cbrnr, line 141: How does the history reflect the inversion of the mappings dict? It should somehow, otherwise copy and pasting the history won't work as a script.

mnelab/model.py
```diff
@@ -131,22 +131,39 @@ def find_events(self, stim_channel, consecutive=True, initial_event=True,
...
+            annots = mne.annotations_from_events(
+                self.current["events"],
+                self.current["data"].info["sfreq"],
+                event_desc=self.current["event_mappings"]
+            )
+        else:
+            annots = mne.annotations_from_events(
+                self.current["events"],
+                self.current["data"].info["sfreq"],
+            )
```
cbrnr, line 156: This if/else block could be shortened because the only difference is the `event_desc` argument, which you could set to `None` in case it is not available and then you can use one function call. Same for the block further down in line 159.

mnelab/model.py
abcsds: You're right. I missed it. Thanks!

mnelab/model.py
abcsds: Thanks for the suggestion! I hadn't thought about it. I followed the example of what you did on the `plot_data` function.

mnelab/model.py
```diff
@@ -131,22 +131,36 @@ def find_events(self, stim_channel, consecutive=True, initial_event=True,
...
     def annotations_from_events(self):
         """Convert events to annotations."""
-        annots = mne.annotations_from_events(
-            self.current["events"],
-            self.current["data"].info["sfreq"]
-        )
+        if "event_mappings" in self.current:
+            mappings = self.current["event_mappings"]
+        else:
+            mappings = None
```
cbrnr, line 150: One more optimization, this can be shortened to `mappings = self.current.get("event_mappings")`.

mnelab/model.py
```diff
@@ -131,22 +131,36 @@ def find_events(self, stim_channel, consecutive=True, initial_event=True,
...
+        annots = mne.annotations_from_events(self.current["events"],
+                                             self.current["data"].info["sfreq"],
+                                             event_desc=mappings
+                                             )
```
cbrnr, line 154: You can move the `)` to the end of the previous line.

mnelab/model.py
```diff
@@ -131,22 +131,36 @@ def find_events(self, stim_channel, consecutive=True, initial_event=True,
...
-        events, _ = mne.events_from_annotations(self.current["data"])
+        events, mappings = mne.events_from_annotations(self.current["data"])
+        # Mappings for annots are swapped {str: int} should be {int: str}
+        mappings = dict((v, k) for k, v in mappings.items())
         if events.shape[0] > 0:
             self.current["events"] = events
-            self.history.append("events, _ = "
-                                "mne.events_from_annotations(data)")
+            self.current["event_mappings"] = mappings
+            self.history.append("events, event_mappings = "
```
cbrnr, line 140: Should be `self.history.append("events, mappings = "` I think?

mnelab/model.py
```diff
@@ -131,22 +131,36 @@ def find_events(self, stim_channel, consecutive=True, initial_event=True,
...
+            mappings = None
+        annots = mne.annotations_from_events(self.current["events"],
+                                             self.current["data"].info["sfreq"],
+                                             event_desc=mappings
+                                             )
         if len(annots) > 0:
             self.current["data"].set_annotations(annots)
-            self.history.append("annots = mne.events_from_annotations(data)")
+            hist = ('annots = mne.annotations_from_events(data,'
+                    'data.info["sfreq"],')
```
cbrnr, line 158: Remove the trailing `,` from the string.

mnelab/model.py
```diff
@@ -131,22 +131,36 @@ def find_events(self, stim_channel, consecutive=True, initial_event=True,
...
+            if "event_mappings" in self.current:
+                hist += "event_mappings)"
```
cbrnr, line 160: Insert the comma here: `hist += ", mappings"`

mnelab/model.py
```diff
@@ -131,22 +131,36 @@ def find_events(self, stim_channel, consecutive=True, initial_event=True,
...
+            else:
+                hist += ')'
```
cbrnr, line 162: No `else` needed because the closing paren has to be inserted in any case.

mnelab/model.py
cbrnr: And I guess there needs to be a history entry defining `mappings`...

mnelab/model.py
abcsds: I see. `self.current["variable"]` is only for assignment, right? I updated it.

mnelab/model.py
abcsds: Sorry, I was trying to keep different names for the function variable, the instance attribute name, and the history variable, but it's unclear. I've updated for a concise name across the three, there is no naming collisions, because they refer to a variable, an attribute, and a history string.

cbrnr: I think the dict key `"event_mappings"` is OK, but I'd vote for the shorter name `mappings` both in the code as well as in the history.

cbrnr: Otherwise I think this is ready, so please also add a changelog entry and we're good to go.

mnelab/model.py
cbrnr: No, the `dict.get` method automatically returns `None` if the key is not present, so there's no need to check for `None`.

cbrnr:
I've pushed some updates and added the changelog entry. Thanks @abcsds!

Issue #159, opened 2020-09-01
Replace mnelab.io.read_raw with mne.io.read_raw


cbrnr:
Implemented in 0.21 - to be released September 15.

cbrnr:
Although MNE now has a `mne.io.read_raw` function, it doesn't support XDF yet, so we will have to keep our own `mnelab.io.read_raw` function around for a little longer.

PR #171, opened 2020-09-25
Update to mne.io.read_raw (#159)

abcsds:
Swapped the internal `read_raw` to `mne.io.read_raw` available from mne v0.21.0. Updated history to reflect changes.


cbrnr:
Nice! I think we also need to require MNE >= 0.21 in various places such as README.md.

abcsds:
There is no version requirements in the `requirements.txt` I'm keeping it like that and adding the new requirement only to the `setup.py`.

cbrnr:
Yes, `requirements.txt` doesn't contain version restrictions because it should always get the latest available versions. I use it only for the CIs. There is another `requirements-oldest.txt` which tests the oldest supported versions, so we should upgrade `mne==0.20.0` to `mne==0.21.0`. I really wonder if and how we should maintain this list of oldest supported versions. My focus is on getting MNELAB to work with current versions of package dependencies, so as long as we test that MNELAB works with the latest versions we could also get rid of testing with old versions.

cbrnr:
Furthermore, `mnelab.io.read_raw` can be removed completely now.

cbrnr:
Also, I was wondering how we should now deal with XDF files - `mne.io.read_raw` doesn't support XDF, that's why I manually added support to MNELAB's own `read_raw`...

cbrnr:
The required packages and versions are now located only in `requirements.txt` (#172) - if you rebase please make sure to follow the new format.

abcsds:
Yeah, as long as MNE doesn't have support for XDF files it's probably better to keep `mnelab.io.read_raw`.

PR #173, opened 2020-10-16
Feature bipolar references

gdolle:
- Add a button to set bipolar references (longitudinal)
- Fix args for exported functions "export_files"


cbrnr:
Thanks for your PR! Would you mind splitting it up so that one PR tackles exactly one issue?

Regarding the arguments, can you explain the issue with the current implementation and how your fix works?

Regarding the bipolar reference feature, currently you hard-code the derivations, but I'd prefer that the user decides which channels are anodes and which channels are cathodes. Based on this input, custom bipolar derivations could be computed.

Let me know what you think and if you need help implementing these changes!

gdolle:
> Thanks for your PR! Would you mind splitting it up so that one PR tackles exactly one issue?

Yes ok I'll split in 2 PR. 

> 
> Regarding the arguments, can you explain the issue with the current implementation and how your fix works?

I will explain in the specific new PR (and add a specific issue also)

> 
> Regarding the bipolar reference feature, currently you hard-code the derivations, but I'd prefer that the user decides which channels are anodes and which channels are cathodes. Based on this input, custom bipolar derivations could be computed.

I agree with you. Also maybe add a load/save button to create bipolar reference profile (to avoid setting every time)
I'll have a look (@Qdelannoy)

> 
> Let me know what you think and if you need help implementing these changes!


> Regarding the bipolar reference feature, currently you hard-code the derivations

gdolle:
@cbrnr I updated this PR to select the channels via a dialog, you can give it a try ?

cbrnr:
I'm sorry that I completely overlooked this PR. Now that the codebase has changed a lot I think it's pretty difficult to make this work, but I will add support for bipolar montages in another PR soon!

Issue #176, opened 2020-10-23
Bug with export function call and args number


gdolle:
Hi,

There's a problem with this line that makes export crash
https://github.com/cbrnr/mnelab/blob/master/mnelab/mainwindow.py#L447

It assumes that all filter functions have the same number of arguments which is not the case, for instance

https://github.com/cbrnr/mnelab/blob/master/mnelab/model.py#L162
https://github.com/cbrnr/mnelab/blob/master/mnelab/model.py#L178

I'll propose a PR (@cbrnr ;) )

cbrnr:
Fixed by #177.

PR #177, opened 2020-10-23
Fix export

gdolle:
This PR should solve issue #176

### DIFF
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -28,6 +28,7 @@ from .model import LabelsNotFoundError, InvalidAnnotationsError
 from .utils import have, has_locations, image_path, interface_style
 from .io import writers
 from .io.xdf import get_xml, get_streams
+from inspect import signature
 
 
 __version__ = "0.6.0.dev0"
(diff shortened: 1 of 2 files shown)


cbrnr:
Can you give an example where the current implementation gives an error? `ffilter` has a default value of `"*"`, so I don't know what the problem is.

gdolle:
try to export file->export

cbrnr:
Ah OK!  Hmm, when did I break this? It seems like the export mechanism is overly complicated, I have to figure out what all this extension and ffilter stuff does and if it is really necessary.

gdolle:
I agree it may be cleaner. (I suggest just a workaround) ;)

cbrnr:
Yes, at some point the exporting functionality should be refactored, but for now we need a workaround. Would it be simpler to just add the `ffilter` parameter to the functions in `model.py` that only have an `fname` parameter?

gdolle:
I'm not sure it is better in terms of design.
If more arguments are required later, that mean you have to change the prototype for all your export functions (with unused args)

cbrnr:
I've created #178 to discuss this.

mnelab/mainwindow.py
```diff
@@ -443,8 +444,15 @@ def open_file(self, f, text, ffilter="*"):
...
+        sig = signature(f)
+
+        args={"fname": fname}
+        for i in sig.parameters.items():
+            if(i[0]=="ffilter"):
+                args[i]=ffilter
+
         if fname:
-            f(fname, ffilter)
+            f(*args.values())
```
cbrnr, line 455: I think this could be shortened to:

```python
        if fname:
            try:
                f(fname, ffilter)
            except TypeError:
                f(fname)
```

cbrnr:
Thanks @gdolle!

PR #189, opened 2020-12-11
read_raw_xdf uses first EEG stream, does not require stream_id

lokinou:
Hello Clemens, I did install mnelab as requested :)
The command to discover the stream_id can be complicated, so I tweaked the code to make it optional

If no stream_id given or found, it will instead use the first EEG stream.

### DIFF
--- a/mnelab/io/xdf.py
+++ b/mnelab/io/xdf.py
@@ -6,7 +6,7 @@ import numpy as np
 import mne
 
 
-def read_raw_xdf(fname, stream_id, *args, **kwargs):
+def read_raw_xdf(fname, stream_id=None, *args, **kwargs):
     """Read XDF file.
 
     Parameters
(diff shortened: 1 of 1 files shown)


cbrnr:
So is this related to the annotations problem you mentioned in the gist?

lokinou:
No, the gist dimension problem seems to be gone in this implementation.

Here, the change here seemed appropriate because the `stream_id` for EEG in my xdf change every time (they get scrambled when updating LSL streams in LabRecorder)
It seems unintuitive to call `stream_ids = match_streaminfos(resolve_streams(fname), [{"type": "EEG"}])` 
before being able to call `read_raw_xdf()`.
So I made the stream_id optional and automatically scans for an EEG stream.
I assumed it would not interfere with the current design.

cbrnr:
Great to hear that the annotations issue has been resolved. Thank you also for your PR, but since MNELAB always gets the stream ID via a dialog window, I don't want to integrate these changes. I hope you're not upset, but I didn't really foresee this function to be used in a script - I just thought it will probably work, but you make a good point that currently it's not ideal. I'm closing this for now, but once we start getting this into base MNE (not MNELAB), I will come back to your suggestion (I'll ping you so you get credit for this contribution)!

lokinou:
No offense taken, what prevails is the consistency of mnelab.
If it can help you for later I tried to ensure my changes did not interfere
with the current way this function works.

Thanks for taking a long time explaining it all.
Keep on the great work
Loic

Clemens Brunner <notifications@github.com> schrieb am Sa., 12. Dez. 2020,
12:13:

Issue #195, opened 2021-02-16
Load 2D arrays from .mat and .npy


jgcaffari1:
It would be useful to be able to load ieeg/eeg recordings saved as numpy arrays.  

Is anyone working on this?

cbrnr:
In principle this would be no problem at all, but .npy files can contain arbitrary NumPy arrays. Is there any particular format/device that specifies the structure contained in such a file? Do you have an example file that you could share?

jgcaffari1:
I was thinking about giving it an array with dimensions [channels, time points], converting it to a RawArray, then plotting it with row numbers as lead names.

cbrnr:
I think it would still be useful to have an option to load 2D array data, which could be stored as .npy or probably even more popular .mat. Missing meta-information such as fs could be queried in a dialog.

PR #205, opened 2021-02-26
Added support for .mat and .npy files.

jgcaffari1:
addresses [this enhancement](https://github.com/cbrnr/mnelab/issues/195).

### DIFF
--- a/mnelab/dialogs/__init__.py
+++ b/mnelab/dialogs/__init__.py
@@ -21,10 +21,12 @@ from .pickchannelsdialog import PickChannelsDialog
 from .referencedialog import ReferenceDialog
 from .runicadialog import RunICADialog
 from .xdfstreamsdialog import XDFStreamsDialog
+from .npydialog import NpyDialog
 
 
 __all__ = [AnnotationsDialog, AppendDialog, CalcDialog,
            ChannelPropertiesDialog, CropDialog, EpochDialog, ERDSDialog,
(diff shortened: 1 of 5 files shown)


cbrnr:
Alright, sorry for taking so long to review this. Would you mind implementing support for only NPY files in this PR? I think MAT files require more infos to be specified by the user because they contain several variables, whereas NPY files contain exactly one array. Therefore, I'd tackle MAT import in a separate PR.

Regarding NPY import, I think there should only be two options that users need to set before importing:

1. Sampling frequency
2. Transpose (a checkbox)

Re 1, I usually use a double spinbox widget. I would restrict the values to be greater than 0 instead of 100Hz. I wouldn't impose any other (upper) limit either. Re 2, here it would be necessary to peak into the file to determine the shape of the saved array. The shapes should be displayed in the dialog, and users should be able to transpose the data before importing with a checkbox (by default it should be unchecked).

If you don't have time, we could even add the transpose option at a later time. Then, this PR could be simplified a lot, because the dialog would contain just the sampling frequency. I'd also remove channel type and normalization, because these operations can be done later after the file has been imported.

cbrnr:
Continued at #213.

jgcaffari1:
I am still working on this, I had something come up at work.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

cbrnr:
I already made the changes, please comment in #213.

PR #208, opened 2021-03-20
Add theme detection for Windows

cbrnr:
Currently light/dark theme detection works only on macOS. On Windows and Linux, Qt apps still don't have native dark theme support 😞.

### DIFF
--- a/mnelab/utils/utils.py
+++ b/mnelab/utils/utils.py
@@ -34,9 +34,14 @@ def interface_style():
             return "light"
     elif sys.platform.startswith("win"):
         from winreg import OpenKey, QueryValueEx, HKEY_CURRENT_USER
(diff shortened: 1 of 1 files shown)


cbrnr:
Alright, detection works now, but there are two problems:

1. Qt doesn't detect theme changes on the fly, only on application start. It seems like there's no `QEvent.PaletteChange` emitted on Windows.
2. It seems like there is no built-in Qt dark theme on Windows, because only the icons change with the app theme.

😞

cbrnr:
We could give the Fusion style a shot (on Windows at least), maybe light/dark modes work with that style (https://www.qt.io/blog/dark-mode-on-windows-11-with-qt-6.5).

cbrnr:
Fixed by #398.

PR #214, opened 2021-07-30
Use new fNIRS name type

rob-luke:
This was changed in MNE-Python in https://github.com/mne-tools/mne-python/pull/7923

### DIFF
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -359,9 +359,11 @@ class MainWindow(QMainWindow):
                 enabled and self.model.current["ftype"] in ["XDF", "XDFZ", "XDF.GZ"]
             )
             self.actions["convert_od"].setEnabled(
-                len(mne.pick_types(
-                    self.model.current["data"].info, fnirs="fnirs_cw_amplitude")
+                len(
+                    mne.pick_types(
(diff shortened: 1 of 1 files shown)


mnelab/mainwindow.py
```diff
@@ -360,7 +360,9 @@ def data_changed(self):
                 enabled and self.model.current["ftype"] in ["XDF", "XDFZ", "XDF.GZ"]
             )
             self.actions["convert_od"].setEnabled(
-                len(mne.pick_types(self.model.current["data"].info, fnirs="fnirs_raw"))
+                len(mne.pick_types(
+                    self.model.current["data"].info,fnirs="fnirs_cw_amplitude")
+                )
```
rob-luke, line 365: Flake didn't like my first push. I tried to copy your style here. I can modify as you want.

cbrnr:
Style seems to be OK, but the closing parenthesis looks a bit weird. I wonder why it didn't like the previous version. I'll play around with it and see if flake needs to be adapted. BTW, the errors are all unrelated (but I really need to fix them).

rob-luke:
Yeah I panicked when I saw the errors. I think the double indented bracket looks weird too. Change as you wish and Ill follow your style in future PRs

cbrnr:
Thanks @rob-luke!

PR #215, opened 2021-07-30
Add support for SNIRF files

rob-luke:
As the title implies 😉


cbrnr:
@rob-luke can you rebase please? I think the CI errors should be gone then.

cbrnr:
Thanks @rob-luke!

PR #233, opened 2021-10-29
Support Python 3.8, 3.9, and 3.10

cbrnr:
Remove Python 3.6 and 3.7 and add Python 3.10 support.


cbrnr:
Matplotlib wheels for 3.10 are still missing, but I guess it's now safe to merge because (1) building from source doesn't take that long, and (2) binary wheels will be available very soon anyway.

PR #237, opened 2021-11-16
Replace QtPy with PySide6

cbrnr:
PySide6 is the official Python package so I think it makes sense to use it.


cbrnr:
Qt5 seems to have a weird bug with font rendering. I don't know when this was introduced, but the text "F.1" is printed with the period inside the "F":
Screen Shot 2021-12-15 at 16 31 14
This happens with both `PySide2` and `PyQt5`, so I guess it's really about time to switch to Qt6.

cbrnr:
The issue on the Linux image can be resolved as mentioned in #250. This is really just because the Ubuntu image only contains the most essential packages, so normal (desktop) Ubuntu users won't have to install anything; `pip install mnelab` will work just like on Windows and macOS. Anaconda users can also install with `pip`, which is necessary because there is no recipe for `PySide6` (or `PyQt6` for that matter, and I doubt that there will be one in the near future given that they are still stuck with an old PyQt5 version).

PR #238, opened 2021-11-16
Replace QtPy with PyQt6

cbrnr:
Same as #237 but using PyQt6 instead of PySide6.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,9 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
     - name: Build wheel
-      run: pip wheel -w ~/.wheelhouse .
+      run: |
+        pip install PyQt6-Qt6
+        pip wheel -w ~/.wheelhouse .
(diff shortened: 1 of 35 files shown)


cbrnr:
I have no idea what's going on. @GuillaumeFavelier do you know by any chance why there is an error installing `PyQt6-Qt6` when the package is [available on PyPI](https://pypi.org/project/PyQt6-Qt6/)?

GuillaumeFavelier:
Are you talking about the failure of the `build-wheel` job? I'm not sure either.

Could it be because `python` is not setup? Or the wrong version is installed by default? I do not recall which one is installed on Github Actions `ubuntu-latest` runners.

What happens if you add someting similar to the `test` job before the installation:

```
    - uses: actions/setup-python@v2
      with:
        python-version: ${{ matrix.python-version }}
```

cbrnr:
Closing because PyQt6 has a ton of differences to Qt6 and PySide6 (especially the enums in `Qt` or `QEvent`), which makes it really hard to look up things in the official docs. I think the better option is to go with PySide6 instead.

PR #250, opened 2021-12-16
Test PySide6 on Linux

cbrnr:
Trying to determine the Linux packages needed to run PySide6.


cbrnr:
Now that this works I'll make the required changes in #237.

PR #252, opened 2022-01-03
Fix: Split file extension with py38 compatibility

### LINKED ISSUE #251: Compatibility error: use of "removesuffix" with Python 3.8
Hi ! 
It seems that using string method `removesuffix` in function `split_name_ext` (`mnelab/io/readers.py`) breaks the compatibility with Python 3.8 ( `removesuffix` was introduced in Python 3.9 by PEP 616).
A possible fix is to use `pathlib.Path(fname).with_suffix("")` and convert the result to string.

yop0:
Fixes issue #251.
Previously, loading XDF files with Python 3.8 resulted in an error due to usage of `removesuffix` (Python 3.9+). This PR fixes the problem by using `pathlib.Path.with_suffix`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## [UNRELEASED] - XXXX-XX-XX
-### Fixed 
-- Fix splitting name and extension of XDF files for compatibility with Python 3.8 ([#252](https://github.com/cbrnr/mnelab/pull/252) by [Johan Medrano](https://github.com/yop0))
+### Fixed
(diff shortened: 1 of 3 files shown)


cbrnr:
Thanks for this PR! I forgot that `str.removesuffix` was only added with Python 3.9. Unfortunately, your changes do not work with multiple extensions such as `.xdf.gz`. However, you could mimic `str.removesuffix` with a combination of checking with `str.endswith` and then slicing off the extensions (see e.g. https://stackabuse.com/python-remove-the-prefix-and-suffix-from-a-string/).

yop0:
Thanks for your feedback! I updated to use negative indexing of `len(ext)` as suggested.

cbrnr:
Thanks @yop0!

PR #255, opened 2022-02-05
Make dataset overwrite dialog more explicit

### LINKED ISSUE #103: Make dialog asking to overwrite the current dataset more explicit
The dialog asking to overwrite the current dataset should be more explicit, i.e. it should mention that the alternative to overwriting is creating a new dataset.

hofaflo:
When closed, the dialog falls back to "No", so by default a new dataset is created.
Resolves #103

### DIFF
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -875,15 +875,14 @@ class MainWindow(QMainWindow):
             self.model.duplicate_data()
             return True
         # otherwise ask the user
-        else:
-            msg = QMessageBox.question(
-                self,
-                "Create new data set",
-                "Create new data set?\nIf 'No', the current data set is overwritten.",
-            )
-            if msg == QMessageBox.Yes:  # create new data set
-                self.model.duplicate_data()
-                return True
+        msg = QMessageBox.question(
+            self,
+            "Create new data set",
+            "Create new data set?\nIf 'No', the current data set is overwritten.",
+        )
+        if msg == QMessageBox.Yes:  # create new data set
+            self.model.duplicate_data()
+            return True
         return False
 
     def _add_recent(self, fname):
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -876,10 +876,10 @@ class MainWindow(QMainWindow):
             return True
         # otherwise ask the user
         msg = QMessageBox()
-        msg.setWindowTitle("Create new data set")
-        msg.setText("Create new data set?")
+        msg.setWindowTitle("Modify data set")
+        msg.setText("You are about to modify the current data set. How do you want to proceed?")  # noqa: E501
         msg.addButton("Create new data set", QMessageBox.YesRole)
-        msg.addButton("Overwrite existing data set", QMessageBox.NoRole)
+        msg.addButton("Overwrite current data set", QMessageBox.NoRole)
         if msg.exec() == 0:  # create new data set (button index 0)
             self.model.duplicate_data()
             return True
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -880,6 +880,7 @@ class MainWindow(QMainWindow):
         msg.setText("You are about to modify the current data set. How do you want to proceed?")  # noqa: E501
         createnew_button = msg.addButton("Create new data set", QMessageBox.YesRole)
         msg.addButton("Overwrite current data set", QMessageBox.NoRole)
+        msg.setDefaultButton(createnew_button)
         msg.exec()
         if msg.clickedButton() == createnew_button:
             self.model.duplicate_data()
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -875,12 +875,12 @@ class MainWindow(QMainWindow):
             self.model.duplicate_data()
             return True
         # otherwise ask the user
-        msg = QMessageBox.question(
-            self,
-            "Create new data set",
-            "Create new data set?\nIf 'No', the current data set is overwritten.",
-        )
-        if msg == QMessageBox.Yes:  # create new data set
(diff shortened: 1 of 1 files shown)


cbrnr:
Or what about we make two buttons instead of Yes and No and label them "Create new dataset" and "Overwrite existing dataset"? The default should be "Create new dataset".

hofaflo:
Sure, it's just a bit repetitive 😅 
image

cbrnr:
🤣  true! How about something like this: "You are about to modify the current dataset. How do you want to proceed?" I can't think of a good title RN, but we should be able to come up with something. Do we write "dataset" or "data set" (we should be consistent)? Also, I would say "Overwrite current dataset".

hofaflo:
"Modifying dataset..." as a title maybe?

I'd use "dataset", although "data set" is more common in the code base right now.

cbrnr:
Let's use "Modify dataset" then.  I have no preference regarding data set vs. dataset, but we should be consistent. If the code base uses data set  then maybe let's stick with it (and change any inconsistent occurrences).

hofaflo:
Done

Okay, I'll create a PR to consistently use "data set" once this is merged.

mnelab/mainwindow.py
```diff
@@ -875,12 +875,14 @@ def auto_duplicate(self):
...
-                                       "Overwrite existing data set?")
-            if msg == QMessageBox.No:  # create new data set
-                self.model.duplicate_data()
-                return True
+        msg = QMessageBox()
+        msg.setWindowTitle("Modify data set")
+        msg.setText("You are about to modify the current data set. How do you want to proceed?")  # noqa: E501
+        msg.addButton("Create new data set", QMessageBox.YesRole)
+        msg.addButton("Overwrite current data set", QMessageBox.NoRole)
+        if msg.exec() == 0:  # create new data set (button index 0)
```
cbrnr, line 883: Instead of relying on indexes, would [this approach](https://doc.qt.io/qt-6/qmessagebox.html#advanced-usage) be better?

Also, what is the default now? I think it should be "Create new data set".

I think we should also disable the Esc key for now, because there is no Cancel option really. Cancelling should do nothing and basically not perform the requested action at all. However, this requires a little rework of the current workflow.

hofaflo: > Instead of relying on indexes, would [this approach](https://doc.qt.io/qt-6/qmessagebox.html#advanced-usage) be better?

oh this is nice, thanks!

> Also, what is the default now? I think it should be "Create new data set".

It is "Create new data set."

> I think we should also disable the Esc key for now, because there is no Cancel option really. 

I couldn't find a way to disable <kbd>Esc</kbd>, do you know how this can be done?

> Cancelling should do nothing and basically not perform the requested action at all. However, this requires a little rework of the current workflow.

Maybe all the dialogs for actions which can create or overwrite a dataset could have a checkbox "Overwrite current data set" instead of the extra dialog?

cbrnr: > I couldn't find a way to disable <kbd>Esc</kbd>, do you know how this can be done?

Hm, not sure - can you pass `None` to [`setEscapeButton`](https://doc.qt.io/qt-6/qmessagebox.html#setEscapeButton)?

cbrnr: If this doesn't work I think we have to role our own `QDialog` object and [overwrite the `keyPressEvent`](https://stackoverflow.com/a/3127786/1112283). It probably also works by overwriting the method for our existing `msg`. But we don't need to do it now if it's too much work.

cbrnr: > Maybe all the dialogs for actions which can create or overwrite a dataset could have a checkbox "Overwrite current data set" instead of the extra dialog?

Yes, I don't know, this would mean that we need to add that checkbox to almost all dialogs. I would prefer a separate dialog. But we could delay the action until the second dialog has been accepted (either with creating a new data set or overwriting the current one); if it gets rejected (cancelled) then we don't perform the action.

hofaflo: > Hm, not sure - can you pass `None` to [`setEscapeButton`](https://doc.qt.io/qt-6/qmessagebox.html#setEscapeButton)?

doesn't help :/


> Yes, I don't know, this would mean that we need to add that checkbox to almost all dialogs. 

I guess this could be simplified by creating a mixin which some dialogs derive from 🤔 

> But we could delay the action until the second dialog has been accepted (either with creating a new data set or overwriting the current one); if it gets rejected (cancelled) then we don't perform the action.

That would require to check for a return from (or exception raised by) `auto_duplicate` in those functions, right?

cbrnr:
On macOS the dialog doesn't look very nice (everything is bold):

Screen Shot 2022-02-06 at 10 10 26

In addition, the default is "Overwrite". I think we should create our own little dialog derived from `QDialog`.

cbrnr:
Or maybe the text is OK for now, but the default should be "Create new data set".

cbrnr:
Thanks @hofaflo!

hofaflo:
> On macOS the dialog doesn't look very nice (everything is bold): 

Hmm, that appears to be the default for QMessageBox on macOS ([link](https://bugreports.qt.io/browse/QTBUG-46913)), could you try if placing `msgBox.setStyleSheet("QMessageBox { font-weight: normal; }")` before `msg.exec()` fixes this?

> In addition, the default is "Overwrite".

Good to know, on Windows it's "Create new" (due to the `YesRole`) 🤔

cbrnr:
> Hmm, that appears to be the default for QMessageBox on macOS ([link](https://bugreports.qt.io/browse/QTBUG-46913)), could you try if placing `msg.setStyleSheet("QMessageBox { font-weight: normal; }")` before `msg.exec()` fixes this?

No, this doesn't do anything. Anyway, it's OK for now. If there is time we can create a nicer dialog later.

PR #256, opened 2022-02-05
Add plotting evoked potentials

hofaflo:
This adds support for plotting evoked potentials from "Epochs" datasets.
Right now the "Evoked" object created by "Epochs.average" is used just for the plot and not stored as a separate dataset.

Example dialog and result for reproducing the [mne-python ERP tutorial](https://mne.tools/stable/auto_tutorials/evoked/30_eeg_erp.html):

image

image


cbrnr:
Nice ❤️ !

> Right now the "Evoked" object created by "Epochs.average" is used just for the plot and not stored as a separate dataset.

Yes, it's like the "Plot ERDS maps" function. We should think about if creating TFR and Epochs objects would be a better alternative (not in this PR though). Here, I'd rename the dialog to "Plot evoked response" if it only creates a plot then.

cbrnr:
Should the dialog allow selecting conditions?

hofaflo:
like this?

image

cbrnr:
Yes! Maybe the list could be a bit shorter (showing 5-6 lines without scrolling should be sufficient).

What other options (from the tutorial) could be added? What about showing the individual channels instead of an average? Adding topoplots at specified time points? Combining an average ERP with GFP at the bottom?

hofaflo:
> Maybe the list could be a bit shorter (showing 5-6 lines without scrolling should be sufficient).

👍 

> What about showing the individual channels instead of an average? 

Sure, although to show which event a plot belongs to would required messing with the figure which [`EvokedArray.plot`](https://mne.tools/stable/generated/mne.EvokedArray.html#mne.EvokedArray.plot) returns :/

> Adding topoplots at specified time points? 

👍

> Combining an average ERP with GFP at the bottom?

The tutorial only shows the combination of individual epochs and GFP ([link](https://mne.tools/stable/auto_tutorials/evoked/30_eeg_erp.html#global-field-power-gfp)), combining average and GFP doesn't appear to be possible in [`plot_compare_evokeds`](https://mne.tools/stable/generated/mne.viz.plot_compare_evokeds.html) 🤔

cbrnr:
>> What about showing the individual channels instead of an average?
>
> Sure, although to show which event a plot belongs to would required messing with the figure which [`EvokedArray.plot`](https://mne.tools/stable/generated/mne.EvokedArray.html#mne.EvokedArray.plot) returns :/

See my answer to the next question.

> The tutorial only shows the combination of individual epochs and GFP ([link](https://mne.tools/stable/auto_tutorials/evoked/30_eeg_erp.html#global-field-power-gfp)), combining average and GFP doesn't appear to be possible in [`plot_compare_evokeds`](https://mne.tools/stable/generated/mne.viz.plot_compare_evokeds.html) 🤔

These are individual channels and not individual epochs. That's also what I meant in my comment before.

hofaflo:
Okay, so I'd add the possibility to create something like this (one figure per event type):
image

where the following things are optional:
- spatial colors
- gfp at the bottom
- topomaps at time points

Is there another type of plot you'd like to include?

cbrnr:
We should think about which plots we would like, and then decide if we want to put it all into one dialog or several. Can we think of a canonical way to build an ERP plot? Or are there incompatible settings that would require separate menu entries?

Here are the parameters we're dealing with:

- Which channels? This is also a bit redundant, because users can pick channels before using this dialog (which could always use all present channels then).
- Which events?
- Average channels (plus optional CI) or show all channels?
- GFP (in addition to ERP or just GFP, not sure if this also works with averaged channels)
- Topomaps at specific time points

It is probably best to implement one dialog per MNE function.

Once we have decided how to proceed, we should not do everything in this PR.

hofaflo:
With one dialog per MNE function, we could have these three possibilities:

- Average channels [+CI], all events in one figure: [`plot_compare_evokeds`](https://mne.tools/dev/generated/mne.viz.plot_compare_evokeds.html#mne.viz.plot_compare_evokeds)
  - Channels (although they can be selected/dropped in a previous step, I think this would be nice to have)
  - Events
  - CI yes/no
  - Epoch averaging mode (mean/median)
  - Channel combination mode (mean/median/std/gfp)
- Individual channels, one figure per event: [`Evoked.plot`](https://mne.tools/stable/generated/mne.Evoked.html#mne.Evoked.plot)
  - Channels
  - Events
  - GFP (yes/no/"only")
  - Spatial colors (yes/no) (only applicable if GFP is not "only")
- Individual channels + topomaps, one figure per event: [`Evoked.plot_joint`](https://mne.tools/stable/generated/mne.Evoked.html#mne.Evoked.plot_joint)
  - Topomap times
  - Channels
  - Events
  - GFP (yes/no/"only")
  - Spatial colors (yes/no) (only applicable if GFP is not "only")

As the bottom two are pretty similar parameter-wise, we could use just one dialog for both?

cbrnr:
This sounds great! I'd combine the last two into one dialog. Am I correct that the first option always produces one figure (even with multiple events selected), whereas the last two produce one figure per event?

hofaflo:
> This sounds great! I'd combine the last two into one dialog. 

Great, then I'll start to draft the dialogs. Do you want to add both in this PR or should the "individual channels" be done in a separate one?

> Am I correct that the first option always produces one figure (even with multiple events selected), whereas the last two produce one figure per event?

Almost, the first option produces one figure per channel type (meg/eeg)

cbrnr:
> Great, then I'll start to draft the dialogs. Do you want to add both in this PR or should the "individual channels" be done in a separate one?

That's up to you. It's probably easier in separate PRs.

> Almost, the first option produces one figure per channel type (meg/eeg)

Makes sense!

hofaflo:
Single-commit force push because I messed up rebasing 🙈 

This implements the first of the three options, here's the dialog:

image

I'd call this menu entry "Evoked (average)...", the other two options would then be under "Evoked (individual)...", but naming suggestions are very welcome!

cbrnr:
> To get an `Evoked` object, `Epochs.average` is called [here](https://github.com/hofaflo/mnelab/commit/095548afe8fcd3cf18e851aab37ab8be16e16665#diff-ea92b193a8d64b1c59aeebcb9d0f88a982171e429db17ad790d347941c2e3055R68), which takes an `average` argument.

Got it. Then let's rename the fields to "Average epochs:" and "Combine channels:"?

mnelab/dialogs/__init__.py
```diff
@@ -18,13 +18,32 @@
 from .metainfodialog import MetaInfoDialog
 from .montagedialog import MontageDialog
 from .pickchannelsdialog import PickChannelsDialog
+from .plotevokedcomparisondialog import PlotEvokedComparisonDialog
```
cbrnr, line 21: The module could be called `plotevokeddialog` and contain both dialogs.

hofaflo: Good idea!

PR #257, opened 2022-02-05
Enable loading multiple files

### LINKED ISSUE #169: Enable loading multiple files
Currently, the open dialog supports selecting only one file. However, I think it would be useful if users could select and load multiple files.

hofaflo:
Resolves #169


cbrnr:
Can you please add a changelog entry?

cbrnr:
Thanks @hofaflo!

PR #258, opened 2022-02-07
Enhance "Set reference"

hofaflo:
Currently, if the data set does not contain the reference channel, rereferencing requires to open "Set reference..." twice:
- first, to add the reference channel
- second, to actually rereference to the desired channel(s)

Example, following [this](https://github.com/cbrnr/bci_event_2021#erders-analysis-with-mnelab) tutorial:
image
image

Here's an idea what the dialog could look like to perform both steps at once (called "Modify reference" to reflect both possible actions):
image

This could be the initial state:
image

And only adding a reference channel, without rereferencing would still be possible:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 ### Added
 - Add support to select and load multiple files at once ([#257](https://github.com/cbrnr/mnelab/pull/257) by [Florian Hofer](https://github.com/hofaflo))
 
+### Changed
+- Simplify rereferencing workflow ([#258](https://github.com/cbrnr/mnelab/pull/258) by [Florian Hofer](https://github.com/hofaflo))
+
 ### Fixed
 - Fix splitting name and extension for compatibility with Python 3.8 ([#252](https://github.com/cbrnr/mnelab/pull/252) by [Johan Medrano](https://github.com/yop0))
 
--- a/mnelab/dialogs/referencedialog.py
+++ b/mnelab/dialogs/referencedialog.py
@@ -2,74 +2,80 @@
 #
 # License: BSD (3-clause)
 
-from PySide6.QtWidgets import (QCheckBox, QDialog, QDialogButtonBox,
-                               QGridLayout, QLineEdit, QRadioButton,
-                               QVBoxLayout)
+from PySide6.QtCore import Qt
+from PySide6.QtWidgets import (QDialog, QDialogButtonBox, QGridLayout,
+                               QGroupBox, QLabel, QLineEdit, QListWidget,
+                               QRadioButton, QVBoxLayout)
 
 
 class ReferenceDialog(QDialog):
-    def __init__(self, parent):
+    def __init__(self, parent, available_channels):
         super().__init__(parent)
-        self.setWindowTitle("Modify reference")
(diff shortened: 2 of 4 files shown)


cbrnr:
Yes, I like that. Some suggestions:

- I would put "Set reference" first and "Add reference" second (and use those names)
- You could try button groups instead of a checkbox with a label for both options. Button groups can also be checkable.

hofaflo:
Thanks for the suggestions, this is much nicer (to look at and to implement :D)
image

If I correctly understand what's happening, in case someone wants to both "Add" and "Set" something, the adding should happen first, right? Therefore I'd also put it first in the dialog 🤔

cbrnr:
Yes, adding a reference channel (all zero) to the data should come before re-referencing to one or more existing electrodes. I thought that because people rarely do that, the option should go last. But you are right, it is probably better when you put it first (even though it is disabled by default).

I was thinking that "Set reference" and "Add reference" are very similar. We could use "Re-reference" instead of "Set reference" to highlight that this operation is applied to *existing* channels, whereas "Add reference" adds a *new* channel.

Do you perform any checks (e.g. adding a new reference requires a new name which is not equal to any existing channels name, and re-referencing requires one or more existing channel names)?

hofaflo:
How about this, for maximum clarity?
image

ad checks: I was thinking mne performs those checks anyway, so mnelab could just show the error messages (and remove the data set duplicate, if applicable), like in [`epoch_data`](https://github.com/cbrnr/mnelab/blob/main/mnelab/mainwindow.py#L793-L802), wdyt?

cbrnr:
> How about this, for maximum clarity?

👍 

> ad checks: I was thinking mne performs those checks anyway, so mnelab could just show the error messages (and remove the data set duplicate, if applicable), like in [`epoch_data`](https://github.com/cbrnr/mnelab/blob/main/mnelab/mainwindow.py#L793-L802), wdyt?

I didn't check, but yes, this would work of course! Theoretically we could use a (multi-selection) list widget that shows all existing channels instead of a text edit.

hofaflo:
> Theoretically we could use a (multi-selection) list widget that shows all existing channels instead of a text edit.

Not sure if that's a valid use case, but this wouldn't work if someone wanted to use a newly added channel as part of the rereferencing selection 🤔

cbrnr:
Well, that's a very rare use case, and I think in such a case it is OK that the users opens the dialog twice.

cbrnr:
I tried adding a reference and re-referencing to an existing channel, but got this error:

```
  File "/Users/clemens/Repositories/mnelab/mnelab/mainwindow.py", line 828, in change_reference
    ref = [c.strip() for c in dialog.reref_channellist.text().split(",")]
AttributeError: 'PySide6.QtWidgets.QListWidget' object has no attribute 'text'
```

hofaflo:
Right, my bad, fixed now!

cbrnr (approved):
Nice! I think this is good to go. Additional thought (maybe another PR): the "average" reference is identical to selecting all channels from the list, so we don't really need it. It's just a shortcut basically, and the most common use cases are either one or two channels or the average over all.

cbrnr:
Thanks @hofaflo!

PR #261, opened 2022-02-08
Add drag and drop reordering to sidebar

### LINKED ISSUE #128: Data set list (left sidebar) should be sortable
It would be nice if we could sort the list of data sets in the left sidebar (e.g. via drag and drop). However, this required #127 to be implemented first.

hofaflo:
I didn't get the `QListView` to allow dropping items only between others, not onto them, so this replaces the sidebar's `QListView` and `QStringListModel` with a `QListWidget`. I'm not sure if this could break something?

The following things work:
- drag and drop reorder (+history)
- data set renaming

@cbrnr, does any missing functionality come to mind?

If not, this fixes #128 :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [UNRELEASED] - XXXX-XX-XX
 ### Added
 - Add support to select and load multiple files at once ([#257](https://github.com/cbrnr/mnelab/pull/257) by [Florian Hofer](https://github.com/hofaflo))
+- Add drag and drop reordering to sidebar ([#260](https://github.com/cbrnr/mnelab/pull/260) by [Florian Hofer](https://github.com/hofaflo))
 
 ### Fixed
 - Fix splitting name and extension for compatibility with Python 3.8 ([#252](https://github.com/cbrnr/mnelab/pull/252) by [Johan Medrano](https://github.com/yop0))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 ## [UNRELEASED] - XXXX-XX-XX
+### Added
+- Add support to select and load multiple files at once ([#257](https://github.com/cbrnr/mnelab/pull/257) by [Florian Hofer](https://github.com/hofaflo))
+
 ### Fixed
 - Fix splitting name and extension for compatibility with Python 3.8 ([#252](https://github.com/cbrnr/mnelab/pull/252) by [Johan Medrano](https://github.com/yop0))
 
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -875,12 +875,16 @@ class MainWindow(QMainWindow):
             self.model.duplicate_data()
             return True
         # otherwise ask the user
-        else:
-            msg = QMessageBox.question(self, "Overwrite existing data set",
-                                       "Overwrite existing data set?")
-            if msg == QMessageBox.No:  # create new data set
-                self.model.duplicate_data()
-                return True
+        msg = QMessageBox()
+        msg.setWindowTitle("Modify data set")
+        msg.setText("You are about to modify the current data set. How do you want to proceed?")  # noqa: E501
+        createnew_button = msg.addButton("Create new data set", QMessageBox.YesRole)
+        msg.addButton("Overwrite current data set", QMessageBox.NoRole)
+        msg.setDefaultButton(createnew_button)
+        msg.exec()
+        if msg.clickedButton() == createnew_button:
+            self.model.duplicate_data()
+            return True
         return False
 
     def _add_recent(self, fname):
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -310,14 +310,14 @@ class MainWindow(QMainWindow):
 
     def _sidebar_move_event(self, parent, start, end, destination, row):
         """
-        Triggered when an data set in the sidebar is moved.
+        Triggered when an item in the sidebar is moved.
 
         Parameters
         ----------
         parent : PySide6.QtCore.QModelIndex
             ???
         start : int
-            The source index of the data set.
+            The source index of the item.
         end : int
             Same as start because the sidebar only allows single selection.
         destination : PySide6.QtCore.QModelIndex
@@ -325,20 +325,7 @@ class MainWindow(QMainWindow):
         row : int
             The target index.
         """
-        # first the data set is copied to the target index
-        self.model.data.insert(row, self.model.data[start])
-        self.model.history.append(f"datasets.insert({row}, datasets[{start}]")
-        # if moved upward, the source index is increased by 1
-        if start > row:
-            start += 1
-        # if moved downward, the new index (after removing the original
-        # item) will be 1 less that the target index
-        else:
-            row -= 1
-        self.model.index = row
-        self.model.data.pop(start)
-        self.model.history.append(f"datasets.pop({start})]")
-        self.data_changed()
+        self.model.move_data_set(start, row)
 
     def data_changed(self):
         # update sidebar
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## [UNRELEASED] - XXXX-XX-XX
 ### Added
 - Add support to select and load multiple files at once ([#257](https://github.com/cbrnr/mnelab/pull/257) by [Florian Hofer](https://github.com/hofaflo))
+- Add drag and drop reordering to sidebar ([#260](https://github.com/cbrnr/mnelab/pull/260) by [Florian Hofer](https://github.com/hofaflo))
```
hofaflo, line 4:
```suggestion
- Add drag and drop reordering to sidebar ([#261](https://github.com/cbrnr/mnelab/pull/261) by [Florian Hofer](https://github.com/hofaflo))
```

cbrnr:
I think a `QListWidget` is fine (and probably also better suited to this simple use case).

Can you use a one-liner in the history to reflect moving a data set, e.g. `datasets[2], datasets[0] = datasets[0], datasets[2]`? Currently, there's also a missing `)` for the `insert` command and an extra `]` for `pop`.

Otherwise I don't think this need any more functionality.

cbrnr:
Maybe this is even worth adding a dedicated method in `Model`, e.g. `Model.swap_data(i, j)` or something?

hofaflo:
Yes, a dedicated method in `Model` makes much more sense 😅 

I see no way of doing this in a one-liner, since it's not a swap but a "move" 🤔

cbrnr:
> Yes, a dedicated method in Model makes much more sense 😅

Cool! I'd name it `move_data` though, because the other methods are already called `insert_data`, `update_data`, `remove_data` etc.

> I see no way of doing this in a one-liner, since it's not a swap but a "move" 🤔

Right, of course.

mnelab/model.py
```diff
@@ -502,3 +502,29 @@ def set_events(self, events):
     @data_changed
     def set_annotations(self, onset, duration, description):
         self.current["data"].set_annotations(mne.Annotations(onset, duration, description))
+
+    @data_changed
+    def move_data_set(self, source, target):
```
hofaflo, line 507:
```suggestion
    def move_data(self, source, target):
```

hofaflo:
> Cool! I'd name it `move_data` though, because the other methods are already called `insert_data`, `update_data`, `remove_data` etc.

Good point, done!

mnelab/mainwindow.py
```diff
@@ -264,15 +264,21 @@ def __init__(self, model):
...
+        self.sidebar = QListWidget()
         self.sidebar.setFrameStyle(QFrame.NoFrame)
         self.sidebar.setFocusPolicy(Qt.NoFocus)
-        self.sidebar.setModel(self.names)
+        self.sidebar.setAcceptDrops(True)
+        self.sidebar.setDragEnabled(True)
+        self.sidebar.setDragDropMode(QListWidget.InternalMove)
+        self.sidebar.setDefaultDropAction(Qt.DropAction.MoveAction)
+        self.sidebar.setEditTriggers(QListWidget.DoubleClicked)
+        self.sidebar.model().rowsMoved.connect(self._sidebar_move_event)
```
cbrnr, line 275: The docs state that this is a private signal, which is not a bad thing per se, but I was wondering if using `beginMoveRows` would be a better idea? Would that also improve the docstring of [`_sidebar_edit_event`](https://doc.qt.io/qt-6/qabstractitemmodel.html#beginMoveRows) (I don't like the two `???` descriptions)?

mnelab/mainwindow.py
```diff
@@ -291,10 +297,44 @@ def __init__(self, model):
...
     def data_changed(self):
         # update sidebar
-        self.names.setStringList(self.model.names)
-        self.sidebar.setCurrentIndex(self.names.index(self.model.index))
+        self.sidebar.clear()
+        self.sidebar.insertItems(0, self.model.names)
+        self.sidebar.setCurrentRow(self.model.index)
+        for it in range(self.sidebar.count()):
+            item = self.sidebar.item(it)
+            item.setFlags(item.flags() | Qt.ItemIsEditable)
```
cbrnr, line 337: Why is this required on every call of this function? Can we make the items editable when they are first inserted?

mnelab/mainwindow.py
hofaflo: `beginMoveRows` isn't a signal, so if I understand correctly this would require to subclass `QAbstractItemModel`?

ad ???: Qt is still too much magic for me :D Are `parent` and `destination` just references to the source an target model of the drag and drop operation, i.e. the sidebar's underlying model?

mnelab/mainwindow.py
hofaflo: The sidebar is cleared and newly filled on every `data_changed`, because it doesn't "know" what happened to the `model`. I found a way to avoid this, but it requires a new connection from the `model` to its `view`, as this doesn't work via the `model.data_changed` decorator.

mnelab/mainwindow.py
cbrnr: Hm, `QListWidget` has a const `model` method (which is a `QAbstractItemModel`). I thought that this could be used, but to be honest I don't really know how. I guess if it works the way you implemented it that's fine.

> Are parent and destination just references to the source an target model of the drag and drop operation,

I'd say yes. Not sure if it's necessary to document all parameters, because this function gets called by a signal and not by our code. I also wonder why `flake8` doesn't complain about unused parameters (maybe the docstring should reflect which parameters are unused).

mnelab/mainwindow.py
cbrnr: What about `self.sidebar.model`?

mnelab/mainwindow.py
hofaflo: > I also wonder why `flake8` doesn't complain about unused parameters

Apparently that's not part of the pure flake8 [error codes](https://flake8.pycqa.org/en/latest/user/error-codes.html), but requires a [plugin](https://github.com/nhoad/flake8-unused-arguments).

> maybe the docstring should reflect which parameters are unused

👍

mnelab/mainwindow.py
hofaflo: > So it's about syncing the entries in the model and the sidebar, right? Are you saying that the previous implementation also cleared everything and then added the new items with `self.names` (a `QStringList`)? Yes, it seems like it.

Yes & yes

> My original comment referred to the repeated `setFlags` call in the loop - is this not possible without a loop?

Apparently not :/

[That's](https://github.com/hofaflo/mnelab/commit/095548afe8fcd3cf18e851aab37ab8be16e16665) what my solution would look like... while it would be nice to not repopulate the sidebar all the time, I'm not sure if it's worth to give up the single connection from `Model` to `MainWindow` (in this direction) for that 🤔

hofaflo: Right now, the methods in `Model` that need to trigger some change in the GUI are decorated with `data_changed`, which calls `Model.view.data_changed()`. That's the only place where a `Model` accesses its `view`. The proposed solution would introduce [two](https://github.com/hofaflo/mnelab/commit/095548afe8fcd3cf18e851aab37ab8be16e16665#diff-ea92b193a8d64b1c59aeebcb9d0f88a982171e429db17ad790d347941c2e3055R57) [new](https://github.com/hofaflo/mnelab/commit/095548afe8fcd3cf18e851aab37ab8be16e16665#diff-ea92b193a8d64b1c59aeebcb9d0f88a982171e429db17ad790d347941c2e3055R68) spots where `view` is accessed. While that isn't per se bad, I was thinking you kept the one-manages-all approach for a reason :D

cbrnr: Ah, now I get it! Yes, I like the separation between the model and the view, even though it is not perfect. So yes, let's not introduce more imperfections for now...

cbrnr:
Thanks @hofaflo!

PR #265, opened 2022-02-09
Simplify imports

hofaflo:
- All `__init__.py` files import their members explicitly (without `*`), so `__all__` is redundant.
- Improve readability of dialog module names by
  - removing the `dialog` suffix
  - inserting underscores between words


cbrnr:
Thanks @hofaflo!

PR #266, opened 2022-02-09
Move XDF info to file menu and add XDF streams details button

hofaflo:
- As the "Show information..." button is only applicable to XDF files, it is moved to the file menu and renamed to "Show XDF meta information". 
- In the XDF stream selection dialog, a "Details..." button is added which also opens the meta info window.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 - Add support to select and load multiple files at once ([#257](https://github.com/cbrnr/mnelab/pull/257) by [Florian Hofer](https://github.com/hofaflo))
 - Add drag and drop reordering to sidebar ([#261](https://github.com/cbrnr/mnelab/pull/261) by [Florian Hofer](https://github.com/hofaflo))
 - Add support for plotting evoked potentials averaged over channels (Plot -> Evoked comparison...) ([#256](https://github.com/cbrnr/mnelab/pull/256) by [Florian Hofer](https://github.com/hofaflo))
+- Add "Details..." button to "Select XDF Stream" dialog ([#266](https://github.com/cbrnr/mnelab/pull/266) by [Florian Hofer](https://github.com/hofaflo))
+
 
 ### Changed
 - Simplify rereferencing workflow ([#258](https://github.com/cbrnr/mnelab/pull/258) by [Florian Hofer](https://github.com/hofaflo))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Changed
 - Simplify rereferencing workflow ([#258](https://github.com/cbrnr/mnelab/pull/258) by [Florian Hofer](https://github.com/hofaflo))
+- Move "Show information..." from toolbar to "File" menu and rename to "Show XDF meta information" ([#266](https://github.com/cbrnr/mnelab/pull/266) by [Florian Hofer](https://github.com/hofaflo))
 
 ### Fixed
 - Fix splitting name and extension for compatibility with Python 3.8 ([#252](https://github.com/cbrnr/mnelab/pull/252) by [Johan Medrano](https://github.com/yop0))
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -3,9 +3,12 @@
 - Add support to select and load multiple files at once ([#257](https://github.com/cbrnr/mnelab/pull/257) by [Florian Hofer](https://github.com/hofaflo))
 - Add drag and drop reordering to sidebar ([#261](https://github.com/cbrnr/mnelab/pull/261) by [Florian Hofer](https://github.com/hofaflo))
 - Add support for plotting evoked potentials averaged over channels (Plot -> Evoked comparison...) ([#256](https://github.com/cbrnr/mnelab/pull/256) by [Florian Hofer](https://github.com/hofaflo))
+- Add "Details..." button to "Select XDF Stream" dialog ([#266](https://github.com/cbrnr/mnelab/pull/266) by [Florian Hofer](https://github.com/hofaflo))
+
 
 ### Changed
 - Simplify rereferencing workflow ([#258](https://github.com/cbrnr/mnelab/pull/258) by [Florian Hofer](https://github.com/hofaflo))
+- Move "Show information..." from toolbar to "File" menu and rename to "Show XDF meta information" ([#266](https://github.com/cbrnr/mnelab/pull/266) by [Florian Hofer](https://github.com/hofaflo))
```
cbrnr, line 11:
```suggestion
- Move "Show information..." from toolbar to "File" menu and rename to "Show XDF meta information..." ([#266](https://github.com/cbrnr/mnelab/pull/266) by [Florian Hofer](https://github.com/hofaflo))
```

mnelab/dialogs/xdf_streams.py
```diff
@@ -70,3 +75,6 @@ def effective_srate(self):
     @property
     def prefix_markers(self):
         return self._prefix_markers.isChecked()
+
+    def details(self):
+        self.parent().xdf_meta_info(self.fname)
```
cbrnr, line 80: Interesting, I never called a method via `parent()`!

mnelab/mainwindow.py
```diff
@@ -143,6 +139,11 @@ def __init__(self, model):
             lambda: self.export_file(model.export_ica, "Export ICA", "*.fif *.fif.gz")
         )
         file_menu.addSeparator()
+        icon = QIcon.fromTheme("meta-info")
+        self.actions["xdf_meta_info"] = file_menu.addAction(
+            "Show XDF meta information",
```
cbrnr, line 144:
```suggestion
            "Show XDF meta information...",
```

CHANGELOG.md
hofaflo: Oh right, that was a conscious decision, but I forgot to mention it: According to [this](https://docs.microsoft.com/en-us/windows/win32/uxguide/cmd-menus#using-ellipses) the ellipsis indicates further options will be presented once the action is selected, which isn't the case here (the "Details..." label is wrong in that case).

mnelab/dialogs/xdf_streams.py
hofaflo: Yeah, I was just about to create an attribute for `parent` in the `__init__` when I realized it's passed to the base class' `__init__`.

CHANGELOG.md
cbrnr: Right, I always used it when an additional dialog pops up. Could you go over all menu entries to make that consistent (maybe in another PR)? We should leave the ellipsis for Preferences... though - that's standard on macOS but the Windows docs advise against it.

I think I added it because "Show XDF chunks..." requires additional inputs and therefore the ellipsis, which is not the case for "Show XDF meta information".

cbrnr: So I guess in this PR the only thing left is change "Details..." to "Details" then?

mnelab/dialogs/xdf_streams.py
```diff
@@ -52,7 +53,11 @@ def __init__(self, parent, rows, selected=None, disabled=None):
         if not disabled:
             self._prefix_markers.setEnabled(False)
         hbox.addWidget(self._prefix_markers)
+        self.details_button = QPushButton("Details...")
```
hofaflo, line 56:
```suggestion
        self.details_button = QPushButton("Details")
```

CHANGELOG.md
hofaflo: Yes, done!

cbrnr:
Thanks @hofaflo!

PR #268, opened 2022-02-09
Show exceptions in an error message box

hofaflo:
As an example, trying to set a montage with channels not in the chosen configuration leads to this ("Show Details..." shows the full traceback):
image


cbrnr:
Very nice addition! How does the traceback look on Windows? On macOS everything is boldface, so it doesn't really look that great (but I think it's OK for now):
Screen Shot 2022-02-09 at 21 13 59

hofaflo:
Windows:
image

I was thinking a monospaced traceback would be nice, but I couldn't get it to work on windows :/

cbrnr:
> I was thinking a monospaced traceback would be nice, but I couldn't get it to work on windows :/

Yeah, same, but for that I think we would have to create a custom dialog anyways (which would then also automatically look better on macOS because we wouldn't use boldface). I'd say this is something for later.

cbrnr:
BTW, why is there an ellipsis on the Details button? 🤦

hofaflo:
No idea 🤦

cbrnr:
Thanks @hofaflo!

PR #271, opened 2022-02-09
Enhance "Set montage..."

hofaflo:
This extends the "Set montage..." dialog to allow specifying `match_case`, `match_alias`, and `on_missing` parameters of `set_montage`. Additionally, locations in the infowidget are displayed as (located_channels/all_channels) instead of Yes/No.

image

### DIFF
--- a/mnelab/dialogs/montage.py
+++ b/mnelab/dialogs/montage.py
@@ -5,6 +5,7 @@
 from mne.channels import make_standard_montage
 from PySide6.QtCore import Qt, Slot
 from PySide6.QtWidgets import (
+    QCheckBox,
     QDialog,
     QDialogButtonBox,
     QHBoxLayout,
@@ -27,6 +28,15 @@ class MontageDialog(QDialog):
                 if self.montages.item(i).data(0) == selected:
                     self.montages.item(i).setSelected(True)
         vbox.addWidget(self.montages)
+
+        self.match_case = QCheckBox("Match case sensitive", self)
+        self.match_case.setChecked(True)
+        self.match_alias = QCheckBox("Match alias", self)
+        self.ignore_missing = QCheckBox("Ignore missing", self)
+        vbox.addWidget(self.match_case)
+        vbox.addWidget(self.match_alias)
+        vbox.addWidget(self.ignore_missing)
+
         hbox = QHBoxLayout()
         self.view_button = QPushButton("View")
         self.view_button.clicked.connect(self.view_montage)
--- a/mnelab/dialogs/montage.py
+++ b/mnelab/dialogs/montage.py
@@ -32,6 +32,7 @@ class MontageDialog(QDialog):
         self.match_case = QCheckBox("Match case-sensitive", self)
         self.match_alias = QCheckBox("Match aliases", self)
         self.ignore_missing = QCheckBox("Ignore missing", self)
+        self.ignore_missing.setChecked(True)
         vbox.addWidget(self.match_case)
         vbox.addWidget(self.match_alias)
         vbox.addWidget(self.ignore_missing)
--- a/mnelab/dialogs/montage.py
+++ b/mnelab/dialogs/montage.py
@@ -30,7 +30,6 @@ class MontageDialog(QDialog):
         vbox.addWidget(self.montages)
 
         self.match_case = QCheckBox("Match case-sensitive", self)
-        self.match_case.setChecked(True)
         self.match_alias = QCheckBox("Match aliases", self)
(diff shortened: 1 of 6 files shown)


cbrnr:
I would uncheck "Match case-sensitive" (please insert a dash) by default. Also, I think "Match aliases" sounds better.

Speaking of case-sensitive, would sorting the entries case-insensitively be better? This could be changed in as part of this PR.

mnelab/dialogs/montage.py
```diff
@@ -4,8 +4,9 @@
 
 from mne.channels import make_standard_montage
 from PySide6.QtCore import Qt, Slot
-from PySide6.QtWidgets import (QDialog, QDialogButtonBox, QHBoxLayout, QListWidget,
-                               QPushButton, QVBoxLayout)
+from PySide6.QtWidgets import (QCheckBox, QDialog, QDialogButtonBox,
+                               QHBoxLayout, QListWidget, QPushButton,
+                               QVBoxLayout)
```
cbrnr, line 9: Please use 92 character lines.

hofaflo:
> I would uncheck "Match case-sensitive"

In [set_montage](https://mne.tools/stable/generated/mne.io.Raw.html#mne.io.Raw.set_montage), `match_case` is `True` by default, should I change this nonetheless?

> Speaking of case-sensitive, would sorting the entries case-insensitively be better? This could be changed in as part of this PR.

👍

cbrnr:
Final nitpick: I'd sort the montages in a "natural" instead of alphabetical way so that we have

```
biosemi16
biosemi32
biosemi64
biosemi128
biosemi160
biosemi256
...
```

For example with this function:

```python
def natural_sort(lst):
    """Sort a list in natural order."""
    def key(s):
        return [int(t) if t.isdigit() else t.lower() for t in re.split(r'(\d+)', str(s))]
    return sorted(lst, key=key)
```

cbrnr:
And maybe there should also be an option to clear a montage/locations (completely). This is not possible with an MNE function (I think), but we could do it manually.

cbrnr:
Clearing locations can be done in another PR.

hofaflo:
Ad natural sorting: very nice idea, implemented in 5422a91

Ad location clearing: That's actually easy with mne (`data.set_montage(None)`), but I'm not sure where this option should best be located 🤔

cbrnr:
Hm, good point. Maybe as a button in "Set montage..."? Clearing is also a kind of setting. That's maybe good enough for now. Alternatively, there could be a separate menu entry called "Clear locations" (or "Clear montage"), but this seems not very elegant.

Speaking of "locations" vs "montage", maybe we should generally use "locations"?

cbrnr:
> Ad clear button: I like the separate menu entry better, so I'd implement that if it's fine with you?

Yes, that's fine, maybe put the two montage entries between two separators.

> Montage: easycap-M1 (32/35 locations)

This would be perfect I think! How do we keep track of the montage? Make a new attribute?

hofaflo:
> Yes, that's fine, maybe put the two montage entries between two separators.

👍 

> This would be perfect I think! How do we keep track of the montage? Make a new attribute?

I'd add an entry to the data set dict, yes!

cbrnr:
"Clear locations" should be disabled if there are no locations currently assigned.

cbrnr:
And I think "Ignore missing" should be checked by default.

mnelab/mainwindow.py
```diff
@@ -617,7 +622,7 @@ def channel_properties(self):
 
     def set_montage(self):
         """Set montage."""
-        montages = mne.channels.get_builtin_montages()
+        montages = natural_sort(mne.channels.get_builtin_montages())
         # TODO: currently it is not possible to remove an existing montage
```
cbrnr, line 626: This comment can be removed.

mnelab/model.py
```diff
@@ -361,9 +367,24 @@ def set_channel_properties(self, bads=None, names=None, types=None):
...
-        self.history.append(f"data.set_montage('{montage}', raise_if_subset=False)")
+    def set_montage(
+        self,
+        montage,
+        match_case=False,
+        match_alias=False,
+        on_missing="raise",
+    ):
+        self.current["data"].set_montage(
+            montage=montage,
```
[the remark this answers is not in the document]
hofaflo: Yes, the location arrays are filled with `np.nan` ([link](https://github.com/mne-tools/mne-python/blob/c3ff05c05c7bfcf765249f27a517803f27f48b4e/mne/channels/montage.py#L917-L919))

cbrnr:
Thanks @hofaflo!

PR #272, opened 2022-02-09
Use short names in sidebar, absolute paths in infowidget

### LINKED ISSUE #260: Data set names in sidebar should not contain full path
I think it would be sufficient if data set names (shown in the sidebar) contained only the base file name (without extension), and not the full path. Not sure if we should disambiguate in case someone loads the same file multiple times (I don't think that's necessary, because users can manually rename if they want).

hofaflo:
Resolves #260 

Example screenshot:
image

This also ensures absolute paths are shown in the infowidget for files loaded via commandline arguments.


cbrnr:
Thanks @hofaflo!

PR #273, opened 2022-02-09
Print caught exceptions to stderr

hofaflo:
If something crashes before the application is able to show the error message box introduced in #268, there's no error message at all. While this should hopefully only happen to developers, it's helpful if all exceptions are also sent to stderr 😅


cbrnr:
Thanks @hofaflo!

PR #275, opened 2022-02-10
Add isort pre-commit hook

hofaflo:
I initially just wanted to add an `isort` key in `setup.cfg` to avoid defining the value in the VSCode settings, then figured we might as well ensure it in pre-commit. Of course this comes with cleaning up all imports at once to make the style checks green 😅


cbrnr:
Perfect! Thanks @hofaflo!

PR #278, opened 2022-02-10
Add ERDS topomap plotting

hofaflo:
To-Do:
- [x] use `dialogs.utils.select_all` once  #277 is merged

There's a lot of code duplication in `dialogs/erds.py`, but since we want to add significance masks to ERDS maps soon, I'd postpone refactoring to when that is done.


cbrnr:
Do you want me to merge this one or is there something you wanted to refactor?

cbrnr:
I just tested this PR and I have three comments:

1. The menu entry should be disabled if there are no locations (montage) associated with the data.
2. What is calculated here? The mean of the frequency and time range? Why is step size important then (for the user)?
3. How are the plots scaled? The colorbar is labeled with "AU", but since we're computing ERDS it should be labeled as such. Also, the lower boundary should be -100% (or -1, I don't even know if we're showing values in percent or not). And the upper boundary is usually between 2.5 and 3.5, maybe this should be settable (with an option to determine the maximum from the data)?

Do you have any good example data where we can test this? Maybe the data from the [ERDS maps tutorial](https://mne.tools/dev/auto_examples/time_frequency/time_frequency_erds.html)? But here the channel labels end with dots and there's no easy way to remove them in MNELAB...

hofaflo:
I'm lacking ERDS knowledge, so I think it would be best if we discuss this on tuesday in "person" 😅 

I used the data from your BCI-event [ERDS Tutorial](https://github.com/cbrnr/bci_event_2021) for trying things out.

hofaflo:
The cmap used in `viz.plot_erds_topomaps` is now the same as in `viz.plot_erds`. #299 will change `viz.plot_erds` to use cnorm. This is not applicable here as [`mne.viz.plot_tfr_topomap`](https://mne.tools/stable/generated/mne.viz.plot_tfr_topomap.html) does not have a cnorm parameter 🤔

cbrnr:
Then I guess we should include `center_cmap` in MNELAB for now. The behavior of `TwoSlopeNorm` has changed in recent Matplotlib versions anyway, the colorbar has a weird scaling which I don't like (see my question [here](https://discourse.matplotlib.org/t/colorbar-scale-for-twoslopenorm/22592)).

cbrnr:
The dialog is also missing edits to select the time range that should be averaged.

hofaflo:
> The dialog is also missing edits to select the time range that should be averaged.

Strange, it's there for me 🤔 
image

cbrnr:
Maybe I forgot to pull. I though VS Code does this automatically.

hofaflo:
The map is always computed with the entire epoch duration (there's no time parameter [here](https://github.com/hofaflo/mnelab/blob/erds-topomap/mnelab/viz.py#L78)). The time range that's entered in the dialog is then applied [here](https://github.com/hofaflo/mnelab/blob/erds-topomap/mnelab/viz.py#L80).

cbrnr:
I see, OK, that makes sense.

cbrnr:
Do you want to include `center_cmap` in this PR?

hofaflo:
> Do you want to include `center_cmap` in this PR?

I'm not sure what you mean 🤔

cbrnr:
Include it in MNELAB because it will be gone from MNE.

cbrnr:
Thanks @hofaflo!

PR #279, opened 2022-02-11
Add significance masks to Plot ERDS maps

hofaflo:
To allow cancelling the significance mask calculation, this required more changes than expected...
The figures created by [`AverageTFR.plot`](https://mne.tools/stable/generated/mne.time_frequency.AverageTFR.html#mne.time_frequency.AverageTFR.plot) can't be pickled (and thus not passed between processes), so we can't just extend the existing `mnelab.viz.plot_erds` with significance masks and call it like it's done for ICA calculation. I've created a new function `calc_tfr_and_masks` which just calculates the tfr and significance masks in a new process. After that, `plot_erds` takes the return of `calc_tfr_and_masks` to create the figures.
Naming and data structure are of course up for discussion 😅

### DIFF
--- a/mnelab/viz.py
+++ b/mnelab/viz.py
@@ -152,7 +152,7 @@ def plot_erds(tfr_and_masks):
         widths = n_cols * [10] + [1]  # each map has width 10, each colorbar width 1
         fig, axes = plt.subplots(n_rows, n_cols + 1, gridspec_kw={"width_ratios": widths})
         vmin, vmax = -1, 2  # default for ERDS maps
-        cmap = center_cmap(plt.cm.RdBu, vmin, vmax)
+        cmap = _center_cmap(plt.cm.RdBu, vmin, vmax)
 
         # skip the last column in `axes`, as it contains the colorbar
         for (ch_name, mask), ax in zip(masks.items(), axes[..., :-1].flat):
--- a/mnelab/viz.py
+++ b/mnelab/viz.py
@@ -6,6 +6,7 @@ import math
 
 import matplotlib.pyplot as plt
 import numpy as np
+from mne.stats import permutation_cluster_1samp_test as pcluster_test
 from mne.time_frequency import tfr_multitaper
 from mne.viz import plot_compare_evokeds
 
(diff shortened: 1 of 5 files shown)


cbrnr:
@hofaflo do you know what's going on with Azure? Somehow it doesn't find an Ubuntu package anymore...

hofaflo:
I only found [this](https://status.dev.azure.com/_event/288504674) and people are reporting similar issues on [twitter](https://twitter.com/AzureDevOps/with_replies) 🤔

cbrnr:
Oh, so all we can do is wait then...

mnelab/viz.py
```diff
@@ -19,30 +21,110 @@ def _get_rows_cols(n):
...
     tfr.apply_baseline(baseline, mode="percent")
     tfr.crop(*times)
 
+    pcluster_kwargs = dict(
+        n_permutations=100,
+        step_down_p=0.05,
+        seed=1,
+        buffer_size=None,
+        out_type='mask',
+    )
```
hofaflo, line 61: Except for alpha, I left the pcluster parameter hardcoded as in the [MNE-Python example](https://mne.tools/stable/auto_examples/time_frequency/time_frequency_erds.html). Do you want to expose some of them?

cbrnr: Hm, I don't think that's necessary, at least for now. Did you see that there is an `n_jobs` parameter? Should we use it and would that work in combination with starting in a separate process?

hofaflo: I did, but didn't think further about it since joblib isn't a (direct) dependency of mnelab at the moment... but will try that!

hofaflo: Apparently not: `UserWarning: Loky-backed parallel loops cannot be called in a multiprocessing, setting n_jobs=1`. Do you want to add that as an option nonetheless?

cbrnr: Nah, if it doesn't work we don't need it. I'd rather have a dialog which can cancel the computation.

cbrnr:
I mean, [they say they have fixed the issue](https://status.dev.azure.com/_event/288504674) but it's still not working...

cbrnr:
I cannot cancel the operation. When I click on "Cancel", I get a spinning beachball and the app doesn't respond. The only way I can revive it is with a keyboard interrupt in the console.

Canceling the ICA calculation does work though. Did you use the same mechanisms? The ICA mechanism doesn't use `QMetaObject.invokeMethods()` for example.

hofaflo:
Oh right, I had added a `res.get()` to get a traceback in case the function in the child process crashes.

The mechanisms should be exactly the same, i copied the `callback` function from the [ICA implementation](https://github.com/hofaflo/mnelab/blob/erds-significance/mnelab/mainwindow.py#L918) 🤔

cbrnr:
Now it works perfectly!

Final comments:

- The Calculating dialog is too narrow, please make it wider so that the title text is displayed. Also, can you please check if the ICA Calculating dialog is non-resizable? If so, this dialog should be as well.
Screen Shot 2022-02-16 at 12 50 16
- Instead of a separate group for "Significance mask", I'd create a checkable label/lineedit labeled "Significance level" (unchecked by default). Alternatively (if you want to keep the group), you can also find a better name than "alpha".
- Regarding naming, I'd use a shorter name `calc_tfr` instead of `calc_tfr_and_masks`. Also, should this function better be private?

hofaflo:
> The Calculating dialog is too narrow, please make it wider so that the title text is displayed. 

Changed, please check if this is wide enough (on Windows there's only the "X" button, so there's more space in the title bar)

> Also, can you please check if the ICA Calculating dialog is non-resizable? If so, this dialog should be as well.

The ICA-dialog is resizable, so I left this one resizable too.

> Instead of a separate group for "Significance mask", I'd create a checkable label/lineedit labeled "Significance level" (unchecked by default). Alternatively (if you want to keep the group), you can also find a better name than "alpha".

Changed to checkbox with label, and replaced the lineedit with a spinbox for a nicer layout:
image

> Regarding naming, I'd use a shorter name `calc_tfr` instead of `calc_tfr_and_masks`. Also, should this function better be private?

Changed x2 :)

cbrnr:
Thanks @hofaflo!

PR #283, opened 2022-02-12
Enable editing events and annotations when there are none

### LINKED ISSUE #274: Edit annotations and Edit events should be enabled
Currently, both "Edit annotations" and "Edit events" menu entries are disabled when a data set is active which doesn't have any annotations or events, respectively. However, I think we should enable both dialogs because they can be used to *add* new annotations/events, which currently isn't possible.

hofaflo:
resolves #274


cbrnr:
Works great! Can you add a changelog entry?

Idea for an additional feature (not for this PR though): Add an option to switch units between samples and seconds. Currently, only samples are possible for both events and annotations, but in MNE-Python annotations are stored in seconds.

hofaflo:
I'll put it on my To-Do list :)

cbrnr:
Thanks @hofaflo!

PR #284, opened 2022-02-12
Allow importing events from FIF files

hofaflo:
To facilitate reproducing [this tutorial](https://mne.tools/stable/auto_tutorials/evoked/30_eeg_erp.html)


cbrnr:
Could we use just `*.fif` for the extension? The `-eve` part is a convention that not everyone uses. If the file doesn't contain any events, there should be an error anyway (but of course we should try).

hofaflo:
Seems to work, FIF files without events give an error (from mne).

cbrnr:
This needs another rebase.

cbrnr:
Can you add a changelog entry please?

hofaflo:
done!

cbrnr:
Thanks @hofaflo!

PR #285, opened 2022-02-12
Add ability to pick channels by channel type

hofaflo:
Example dialog:

image

The existing implementation of `pick_channels` converts the "pick" to a "drop". What was the reasoning there and should it be done in this case as well?


cbrnr:
No specific reason, I couldn't decide what's better from a UX point of view (picking or dropping), and then I probably forgot to change the function. I think they should perform the same thing though, just the inverse. For consistency, maybe it's better to only use pick functions though.

hofaflo:
> For consistency, maybe it's better to only use pick functions though.

I agree, done!

PR #286, opened 2022-02-12
Improve infowidget montage entry

hofaflo:
Fixes for two small issues:
- If the channel locations are included in the loaded file, the key "montage" in the data set dict is not set. As this is used to decide whether the montage details in the infowidget should be shown, they aren't.
- The `nchan` string includes "(X bad)" in case of bad channels.

### DIFF
--- a/mnelab/model.py
+++ b/mnelab/model.py
@@ -312,11 +312,14 @@ class Model:
         if isinstance(reference, list):
             reference = ",".join(reference)
 
-        if montage is None:
+        locations = count_locations(self.current["data"].info)
+
+        if montage is None and not locations:
             montage_text = "-"
-        else:
-            locations = count_locations(self.current["data"].info)
-            montage_text = f"{montage} ({locations}/{nchan} locations)"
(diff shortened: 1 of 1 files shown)


cbrnr:
I don't quite understand. How did you change the case when locations are included in a file? We don't know the montage name in the case, right? If that's the case, should we display "custom" or something?

Did the nchan already include "(X bad)"? Or did you add it?

cbrnr:
I like it! I would show "none" instead of "-", WDYT?

> It already did, that happens [here](https://github.com/cbrnr/mnelab/blob/main/mnelab/model.py#L290).

So what did you change here?

hofaflo:
> I like it! I would show "none" instead of "-", WDYT?

:+1:

> > It already did, that happens [here](https://github.com/cbrnr/mnelab/blob/main/mnelab/model.py#L290).
> 
> So what did you change here?

I directly use `data.info["nchan"]` instead of the local, modified `nchan`

cbrnr:
Thanks @hofaflo!

PR #287, opened 2022-02-12
Add ability to drop bad epochs

hofaflo:
This provides an interface for [mne.Epochs.drop_bad](https://mne.tools/stable/generated/mne.Epochs.html#mne.Epochs.drop_bad).

Example dialog:
image

Only available channel types are displayed, alternatively we could always show all but grey-out non-available ones?

### DIFF
--- a/mnelab/dialogs/__init__.py
+++ b/mnelab/dialogs/__init__.py
@@ -7,6 +7,7 @@ from .append import AppendDialog
 from .calc import CalcDialog
 from .channel_properties import ChannelPropertiesDialog
 from .crop import CropDialog
+from .drop_bad_epochs import DropBadEpochsDialog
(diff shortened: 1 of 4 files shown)


cbrnr:
I think it would be important to indicate the units of the thresholds people are supposed to enter. Maybe this can be shown after the type in parentheses, e.g. "EEG (V)"? The units should be available somewhere in the constants module in MNE.

And yes, I think it's fine if only available types are shown.

hofaflo:
Good idea! E.g. "stim" channels don't have a unit, is "AU" fine for that or do you prefer something else?

image

cbrnr:
This looks great! For no unit channels I'd write nothing, so just `STIM`.

cbrnr:
Thanks @hofaflo!

PR #288, opened 2022-02-12
Fix event import

hofaflo:
Since #283 an array with 0 rows is used to represent no events (instead of None), so the if-condition in `Model.import_events` (line 232) needs to be based on length. Also, the event array has to be of dtype `int` ([ref](https://mne.tools/stable/glossary.html#term-events)).


cbrnr:
Thanks @hofaflo!

PR #289, opened 2022-02-13
Add a settings menu

hofaflo:
It's rather minimal so far, but I'm sure people will think of things to add :D

image

I created a common interface for both the implicit (window size/pos, recent files) and the explicit (max_recent, ?) settings. "Reset to defaults" does however only reset what can be changed in the settings dialog. Should this be extended to the rest? And should toolbar and statusbar be toggleable from the settings dialog?


hofaflo:
> I think so, yes. It can be useful if the window size or position is messed up.

Good point, I added a dedicated button to avoid resizing the window if someone just wants to reset the other settings:

image

cbrnr:
Thanks @hofaflo!

PR #290, opened 2022-02-13
Execute `sudo apt update` before installing linux packages

hofaflo:
Tested in my fork ([link](https://github.com/hofaflo/mnelab/actions/runs/1837378220)), this appears to fix the workflow issues 🤷

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,6 +42,7 @@ jobs:
         name: mnelab-wheel
     - name: Install Linux packages
       run: |
+        sudo apt update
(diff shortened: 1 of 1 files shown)


cbrnr:
🤯 - to be fair, it something like that in a message...

cbrnr:
Could you rebase all open PRs so that we can re-run the jobs?

hofaflo:
> 🤯 - to be fair, it something like that in a message...

True, but blaming the package archive is easiest 😅 

> Could you rebase all open PRs so that we can re-run the jobs?

Sure!

PR #293, opened 2022-02-14
Add a "Cancel" button to the overwrite dialog

hofaflo:
To avoid the extra dialog when modifying a data set, we could ask whether a user wants to overwrite or duplicate a data set directly in the respective action dialog:

image

For data sets directly connected to a file, the "Overwrite" button would be disabled, with a tooltip explanation:

image

I've implemented this only for the filter dialog so far to have an example for detail discussions.

I'm not sure whether inheritance will be the best approach to add this ability to multiple dialogs, but had no better idea either, WDYT @cbrnr?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Add ability to pick channels by channel type ([#285](https://github.com/cbrnr/mnelab/pull/285) by [Florian Hofer](https://github.com/hofaflo))
 - Add a settings menu (File -> Settings...) ([#289](https://github.com/cbrnr/mnelab/pull/289) by [Florian Hofer](https://github.com/hofaflo))
 - Add ability to import events from FIF files ([#284](https://github.com/cbrnr/mnelab/pull/284) by [Florian Hofer](https://github.com/hofaflo))
+- Add a "Cancel" button to the overwrite dialog ([#293](https://github.com/cbrnr/mnelab/pull/293) by [Florian Hofer](https://github.com/hofaflo))
 
 ### Changed
 - Simplify rereferencing workflow ([#258](https://github.com/cbrnr/mnelab/pull/258) by [Florian Hofer](https://github.com/hofaflo))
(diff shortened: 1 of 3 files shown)


cbrnr:
I don't know if I like that better - the dialogs get more complex for sure. I like that there's a Cancel option, could we add that to the existing extra dialog (so that it has three buttons) as an alternative?

hofaflo:
Sure, that could be implemented like [this](https://github.com/hofaflo/mnelab/commit/9fe6e68dec3cd8dc9abcad90433811110d2a68f0). I feel it makes the control flow less intuitive though 🤔

cbrnr:
Maybe this could be a configuration setting 👀? I don't know, this would make everything a lot more complicated than it needs to be. Why do you think an explicit dialog asking what to do with the data is not intuitive?

hofaflo:
> [...] this would make everything a lot more complicated than it needs to be.

For sure, I don't want to implement that 😅 

> Why do you think an explicit dialog asking what to do with the data is not intuitive?

That part is great, but I'm not sure if I would get that clicking "Cancel" on the second dialog actually cancels what was started in the first one. Could be just me though, I'm fine with whatever you prefer!

cbrnr:
We can also cancel the Cancel button 😃 and force the user to do something.

hofaflo:
:D
I didn't manage to disable the close button (X) so far though :/

hofaflo:
Implemented the "Cancel" button, ready for review!

cbrnr:
I assume you tried to disable the Close button with a window flag (https://doc.qt.io/qt-6/qtwidgets-widgets-windowflags-example.html), e.g. as mentioned [here](https://stackoverflow.com/questions/9965224/disable-qdialogs-default-close-button-upper-left-cross-button)?

hofaflo:
Yep, but without success :/

cbrnr:
Sorry, this makes a lot of things more complicated than before. I've tried disabling the close button (it works on macOS at least) and using sensible Escape and default button mappings, can you please test #304? If you still see a Close button on Windows, at least it should map to the "Create new" action.

hofaflo:
Obsolete with #304

PR #297, opened 2022-02-14
Use © symbol in copyright notice

cbrnr:
Hopefully nothing gets confused by using that symbol, but I think it should be OK.


hofaflo:
No crash on windows so far

PR #299, opened 2022-02-15
Replace center_cmap

cbrnr:


cbrnr:
Closing because #278 now includes a `_center_cmap` function definition, which replaces the deprecated `mne.viz.utils.center_cmap`. We still need this function because `mne.viz.plot_tfr_topomap` does not (yet) have a `cnorm` parameter.

PR #301, opened 2022-02-15
Use correct extension for export to .fif.gz

### LINKED ISSUE #300: Exporting data to .FIF.GZ creates incorrect extension
Exporting data to .FIF.GZ creates a file with `.gz.fif.gz` extension.

cbrnr:
Previouly, export to .FIF.GZ exported to a file `<name>.gz.fif.gz`. For some reason, `QFileDialog` appends `.gz` when no extension is given in `<name>`.

This PR removes this extra `.gz` and correctly exports to `<name>.fif.gz`. Note that this means that it is impossible to export to a file named `<name>.gz.fif.gz`, but I think that's OK.

Fixes #300.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 ### Fixed
 - Fix splitting name and extension for compatibility with Python 3.8 ([#252](https://github.com/cbrnr/mnelab/pull/252) by [Johan Medrano](https://github.com/yop0))
 - Fix exporting to BrainVision with annotations starting with "BAD" or "EDGE" ([#276](https://github.com/cbrnr/mnelab/pull/276) by [Clemens Brunner](https://github.com/cbrnr))
+- Exporting to .fif.gz now uses the correct extension on macOS ([#301](https://github.com/cbrnr/mnelab/pull/301) by [Clemens Brunner](https://github.com/cbrnr))
 
 ## [0.7.0] - 2021-12-29
 ### Added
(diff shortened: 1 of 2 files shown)


cbrnr:
@hofaflo WDYT of this change? OK?

hofaflo:
I can't reproduce the original issue, that's what I tried:
- load some data set
- File -> Export data -> FIF.GZ
- enter "test" and press "OK"

The new file is then called "test.fif.gz"

cbrnr:
Strange, might be specific to macOS then. I think I'm going to remove the option to export to .fif.gz, because this is just a zipped .fif (importing will still be possible). WDYT?

cbrnr:
Or we'll leave it in and hope that Qt fixes this bug some time. It's related to the native macOS dialog not correctly handling multiple extensions (the non-native dialog works but looks ugly).

I found these issues that look related:

- https://bugreports.qt.io/browse/QTBUG-44227
- https://bugreports.qt.io/browse/QTBUG-67993

@hofaflo can you confirm that exporting to .fif.gz keeps working as expected on Windows when you use this PR? Try entering just a name (`test`), a name plus some extensions (`test.a.b.c`), and a name plus the expected extensions (`test.fif.gz`).

cbrnr:
I think I found a better solution. When I skip the `filter` argument to `QFileDialog` nothing is added automatically. I think that's the way to go.

PR #302, opened 2022-02-15
Let event selection dialogs show event labels

hofaflo:
This lets dialogs in which users can select events show the event label instead of the integer id. If events aren't generated from annotations, but imported from a file, no event mapping exists, so the ids are shown. A future PR could introduce the possibility to change the event mapping.


hofaflo:
> What about showing both event IDs and labels, e.g.

👍 

> Also, this change is currently only affecting the Epochs dialog. Were you planning to add more changes? If not, then the changelog entry should be modified to reflect what was actually changed.

By passing the mapping to `mne.Epochs`, it's automatically reflected in the plotting dialogs as well.

cbrnr:
Great! Now the only thing I'd add is an option to let users modify the mapping. Maybe this could be done in the "Edit Events" dialog (including a "Clear mapping" button)? Otherwise I fear that people might be wondering where these mappings come from and how to change/edit/clear them.

hofaflo:
done!
image

cbrnr:
Great! I'm sorry I keep making new requests, but this is a really long dialog. I'd put the mapping into its own sub-dialog, accessible through a button "Change mapping..." (to the left in the same row as OK/Cancel). If you don't want to implement it let me know and I'll do it.

hofaflo:
> [...] I'll do it.

Yes, please do!

cbrnr:
Thanks @hofaflo!

PR #303, opened 2022-02-15
Add batch renaming of channels

### LINKED ISSUE #298: Batch rename channels
Currently, channels can be renamed individually in the "Channel properties" dialog. However, sometimes it is necessary to e.g. remove trailing dots from all channels names. A mechanism to do that would be really useful, but I don't know how. [`mne.rename_channels()`](https://mne.tools/stable/generated/mne.rename_channels.html) also takes a callable, but we cannot allow arbitrary functions. Maybe we can offer the most useful ones (e.g. strip things from the beginning or end)?

hofaflo:
Resolves #298, that's what the dialog looks like:

image


cbrnr:
Nice! Should the dialog include a way to pick channels (individually and/or by type), or is this too much?

cbrnr:
I also think that it could be problematic if you allow modifying the beginning and end simultaneously. Wouldn't it be easier if there were four mutually exclusive options?

hofaflo:
That's a great layout idea, thanks! I implemented it and added a preview table:

image

Re channel picking: This might be too confusing 😅

cbrnr:
Thanks @hofaflo!

PR #304, opened 2022-02-16
Improve overwrite dialog

cbrnr:
This PR disables (hides) the close window button in the overwrite dialog. It also makes sure that both the default button and pressing Escape is mapped to "Create new dataset" (so the current data is not accidentally overwritten).
Screen Shot 2022-02-16 at 15 22 53


cbrnr:
@hofaflo can you please test how this behaves on Windows? On macOS I don't see any window control buttons on the dialog anymore. Pressing Escape creates a new dataset. The default button (pressing Enter) also maps to this action.

hofaflo:
Looks and works as intended 🎉
image

PR #311, opened 2022-02-18
Consistent naming

### LINKED ISSUE #295: Use verbs for all actions
I think all actions should at least include a verb (we might consider Windows and macOS UX guidelines, but I'm pretty sure that's the rule). Currently, the following actions (menu entries) do *not* include verbs so we should consider renaming them:

- Edit – Channel properties...
- Edit – Annotations...
- Edit – Events...
- *Everything* in the Plot menu
- Everything in the History menu (not sure this needs verbs though)

hofaflo:
While the main intention is to close #295, i.e. to use verbs in all action names, I took the opportunity to harmonize the names of actions, methods, modules, dialogs, and icons. This should make it easier to find related pieces of code in the future.


mnelab/dialogs/__init__.py
```diff
@@ -2,27 +2,31 @@
...
+    PlotEvokedComparisonDialog,
+    PlotEvokedDialog,
+    PlotEvokedTopomapsDialog,
+)
 from .rename_channels import RenameChannelsDialog
 from .run_ica import RunICADialog
-from .xdf_chunks import XDFChunksDialog
+from .set_montage import SetMontageDialog
+from .show_xdf_chunks import ShowXDFChunksDialog
+from .show_xdf_meta_information import ShowXDFMetaInformationDialog
```
cbrnr, line 31: This is getting too long – can we call it `ShowXDFInformationDialog` everywhere?

mnelab/dialogs/__init__.py
```diff
@@ -2,27 +2,31 @@
...
-from .channel_properties import ChannelPropertiesDialog
-from .crop import CropDialog
+from .change_reference import ChangeReferenceDialog
+from .create_epochs import CreateEpochsDialog
+from .crop_data import CropDataDialog
 from .drop_bad_epochs import DropBadEpochsDialog
-from .epoch import EpochDialog
-from .erds import ERDSDialog, ERDSTopomapsDialog
+from .edit_annotations import EditAnnotationsDialog
+from .edit_channel_properties import EditChannelPropertiesDialog
```
cbrnr, line 12: I'd prefer `ChannelPropertiesDialog` (and `channel_properties.py`).

mnelab/dialogs/__init__.py
```diff
@@ -2,27 +2,31 @@
...
 from .calc import CalcDialog
-from .channel_properties import ChannelPropertiesDialog
-from .crop import CropDialog
+from .change_reference import ChangeReferenceDialog
+from .create_epochs import CreateEpochsDialog
+from .crop_data import CropDataDialog
 from .drop_bad_epochs import DropBadEpochsDialog
-from .epoch import EpochDialog
-from .erds import ERDSDialog, ERDSTopomapsDialog
+from .edit_annotations import EditAnnotationsDialog
```
cbrnr, line 11: I'd prefer `AnnotationsDialog` (and `annotations.py`).

mnelab/dialogs/__init__.py
```diff
@@ -2,27 +2,31 @@
...
+from .edit_channel_properties import EditChannelPropertiesDialog
+from .edit_events import EditEventsDialog
```
cbrnr, line 13: Same here, I'd prefer `EventsDialog` (and `events.py`). I think the reason for keeping these names is that these dialogs are also used for viewing information, not just editing. I know they are in the Edit menu, but we don't have to religiously follow a naming scheme when we have reasons to deviate.

cbrnr:
The documentation menu entry can be globally enabled.

cbrnr:
And feel free to convince me that the comments I made regarding some dialogs/actions are wrong – I am really not sure what would be best.

cbrnr:
Can we start from scratch and only implement the changes in the menus (and the other changes not related to naming)? I'm not really happy with renaming everything...

hofaflo:
> Can we start from scratch and only implement the changes in the menus (and the other changes not related to naming)? I'm not really happy with renaming everything...

Sure, I'll make a new PR containing only those commits!

cbrnr:
Closing in favor of #317.

PR #313, opened 2022-02-19
Add reading multiple XDF streams

hofaflo:


cbrnr:
I tried this with the `clock_offsets.xdf` test file, which consists of 
- 1 EEG stream (19 channels, 1000 Hz)
- 1 EMG stream (1 channel, 256 Hz)
- 1 marker stream (0 Hz)
The EMG starts around 4.7 s, which is visible when viewing the file in SigViewer:
sigviewer
However, in MNELAB it seems like signals that contain `NaN` are not rendered unless the currently visible time segment does _not_ contain any `NaN`s. This means that the EMG channel is not visible in a time segment which includes samples before and after signal onset:
mnelab1
Only if a signal does not contain _any_ `NaN`s in the visible time segment it gets rendered (so skipping forward so that the visible window starts _after_ second 4.7):
mnelab2
I am wondering if this is our fault (because we fill signals that have not yet started with `NaN`s) or if the MNE-Python raw browser does not handle such signals correctly? @hofaflo @larsoner @drammock @marsipu any idea?

cbrnr:
@hofaflo in the stream selection dialog I can still deselect the currently selected stream with <kbd>⌘</kbd> left-click (or <kbd>Ctrl</kbd> left-click on Windows/Linux). This leads to the following error message:

```
  File "/Users/clemens/Repositories/mnelab/mnelab/dialogs/xdf_streams.py", line 128, in _set_suggested_fs
    suggested_fs = max(counts, key=counts.get)
ValueError: max() arg is an empty sequence
```

hofaflo:
> @hofaflo in the stream selection dialog I can still deselect the currently selected stream with ⌘ left-click (or Ctrl left-click on Windows/Linux). This leads to the following error message:
> 
> ```
> Traceback (most recent call last):
>   File "/Users/clemens/Repositories/mnelab/mnelab/dialogs/xdf_streams.py", line 128, in _set_suggested_fs
>     suggested_fs = max(counts, key=counts.get)
> ValueError: max() arg is an empty sequence
> ```

Would you prefer to disable deselecting the last selected row or to avoid the error message and disable the OK button if nothing is selected?

cbrnr:
> Would you prefer to disable deselecting the last selected row or to avoid the error message and disable the OK button if nothing is selected?

I'd prefer forcing at least one selected row. Selecting nothing isn't useful in this dialog.

cbrnr:
It doesn't quite work yet. If I select a second stream, then deselect it again, and then try to deselect the first stream I get the same error (because the only selected row gets deselected).

cbrnr:
Suggestion: we could try to override `selectionChanged()` and ignore it if the new selection is empty?

cbrnr:
I tried this, but it doesn't seem to work. It looks like setting a `QItemSelectionModel` is the recommended solution: https://doc.qt.io/qt-6/qitemselectionmodel.html#details

cbrnr:
If there is no easy solution (which I am almost convinced of now), can you disable the OK button/action if there is no selection? And the `_set_suggested_fs` has to be adapted to deal with an empty selection.

larsoner:
> I am wondering if this is our fault (because we fill signals that have not yet started with NaNs) or if the MNE-Python raw browser does not handle such signals correctly? @hofaflo @larsoner @drammock @marsipu any idea? This happens with both the matplotlib and qt browser backends.

By default we `remove_dc=True`. We hopefully just need to change it to be a `np.nanmean` instead of `np.mean`. Can you check?

cbrnr:
Thanks @larsoner, that was it! https://github.com/mne-tools/mne-python/pull/10392

cbrnr:
Thanks a lot @hofaflo!

PR #314, opened 2022-02-19
Add support for loading MAT files

### LINKED ISSUE #195: Load 2D arrays from .mat and .npy

cbrnr:
2D arrays from arbitrary MAT files can be loaded by selecting the variable and setting the sampling frequency in a dedicated dialog. Fixes #195.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - Add support for plotting ERDS topomaps (Plot -> ERDS topomaps...) ([#278](https://github.com/cbrnr/mnelab/pull/278) by [Florian Hofer](https://github.com/hofaflo))
 - Add possibility to apply significance masks to ERDS plots ([#279](https://github.com/cbrnr/mnelab/pull/279) by [Florian Hofer](https://github.com/hofaflo))
 - Add basic batch renaming of channels "Edit - Rename channels...") ([#303](https://github.com/cbrnr/mnelab/pull/303) by [Florian Hofer](https://github.com/hofaflo))
+- Add support for loading data from .MAT files ([#314](https://github.com/cbrnr/mnelab/pull/314) by [Clemens Brunner](https://github.com/cbrnr))
 
 ### Changed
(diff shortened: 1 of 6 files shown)


cbrnr:
@hofaflo if you have time please test this new feature.

hofaflo:
This works really well, just two notes:

The entries in the "Name" column in the `MatDialog` are indented:
image

Would it make sense to import 3d-arrays as `mne.Epochs`?

cbrnr:
> The entries in the "Name" column in the MatDialog are indented:

I haven't noticed that with data I tested. Which file did you use that has this problem? Did you try a file where the entries are not indented?

> Would it make sense to import 3d-arrays as mne.Epochs?

Definitely. But I think I'll do that at a later point, because this will require juggling more with the dimensions.

hofaflo:
Happens with all files I found, even with a minimal example created like this:
```python
import numpy as np
from scipy.io import savemat

a = np.zeros((100, 2))
savemat("test.mat", {"a": a})
```
image

cbrnr:
Ah, I think that's because I'm displaying that as a `QTreeWidget`. When there's no nesting, this looks like an indentation where normally there would be an arrow. Try loading a .mat file from e.g. data set 1 from http://bnci-horizon-2020.eu/database/data-sets to see how a nested data structure (MATLAB struct) looks like.

Not sure it's worth special-casing when there is no nesting, only if it's not too much effort.

cbrnr:
I'm now showing the file name as the root of the tree so there is no stray indentation. Better?
Screen Shot 2022-02-20 at 10 04 36

cbrnr:
Off-topic: the default icon in the top left corner of a dialog on Windows looks really ugly. Can/should we disable this?

hofaflo:
> I'm now showing the file name as the root of the tree so there is no stray indentation. Better?

Yes that's perfect!

> Off-topic: the default icon in the top left corner of a dialog on Windows looks really ugly. Can/should we disable this?

Not sure if that's worth the effort 😅

cbrnr:
> Not sure if that's worth the effort 😅

If it's easy to do (e.g. set a flag) then probably yes.

PR #281, opened 2022-02-11
Add documentation

hofaflo:
This PR serves to discuss the introduction of documentation pages.

To-Do (links lead to the rendered version of the current state of this PR):
- [x] [Home](https://mnelab--281.org.readthedocs.build/en/281/index.html)
- [x] [Tutorial](https://mnelab--281.org.readthedocs.build/en/281/tutorial/index.html)
  - [x] Installation
  - [x] First steps
- [x] Examples (step-by-step guides for specific use cases)
  - [x] [ERDS](https://mnelab--281.org.readthedocs.build/en/281/examples/erds_maps.html)
  - [x] [ERP](https://mnelab--281.org.readthedocs.build/en/281/examples/erp.html)
- [x] [About](https://mnelab--281.org.readthedocs.build/en/281/about.html)
- [x] link in main readme

Two questions so far:
- What additional information should be included in "first steps"?
- The [MNE-Python tutorial](https://mne.tools/stable/auto_tutorials/evoked/30_eeg_erp.html) for ERPs uses an MEG data set which is part of MNE-sample-data, so it requires downloading a 1.5GB tar.gz. @cbrnr can you suggest an ERP data set (preferably few files and manually downloadable)?

PR #317, opened 2022-02-22
Improve menu

hofaflo:
Only the menu changes from #311, the documentation link in "Help" is now part of #281.


cbrnr:
Thanks! Could you remove the "Meta" from "XDF Meta Information"?

PR #319, opened 2022-02-25
Add app icon

cbrnr:
This needs to be tested on Windows, macOS, and Linux. It might not be possible to make this 100% work without creating a standalone version (see https://doc.qt.io/qt-6/appicon.html), but maybe it can be good enough.

One minor thing I noticed is that when quitting MNELAB on macOS, the default Python rocket icon is briefly visible. But I guess that's not too bad.


hofaflo:
Seems to work on Win10, but the icon is relatively small:
image

cbrnr:
What about now?

cbrnr:
Looks good I think:
Screenshot 2022-02-26 131026

PR #333, opened 2022-03-23
Add complete history for find_events

### LINKED ISSUE #331: History of "Find events" does not include selected arguments
There are several options that can be set in the "Find Events" dialog, but these settings are not reflected in the history entry.

cbrnr:
This adds proper history for the Find Events dialog. The argument `consecutive` now also supports `"increasing"` (the default) in addition to `True` and `False`. Fixes #331.


cbrnr:
Not sure why the tests don't run...

Issue #305, opened 2022-02-18
Add "close" button for data sets in sidebar


hofaflo:
The sidebar could show a "close" buttons (X) next to each data set. To avoid unintentional closing, a confirmation dialog should pop up.

PR #342, opened 2022-03-31
WIP,ENH: sidebar context menu

GuillaumeFavelier:
This PR shows a basic context menu when the user right-clicks on an element of the sidebar. It's another take on the suggestion described in https://github.com/cbrnr/mnelab/issues/305. Plus other useful actions could eventually be added to the context menu:

https://user-images.githubusercontent.com/18143289/161091797-35826969-5a28-4fb1-936a-d11e9ffbecc7.mp4

### DIFF
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -320,6 +320,9 @@ class MainWindow(QMainWindow):
             self.toolbar.hide()
             self.actions["toolbar"].setChecked(False)
 
+        # set the actions of the context menu (right click on sidebar)
+        context_menu_actions = [self.actions["close_file"]]
+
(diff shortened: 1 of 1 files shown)


GuillaumeFavelier:
> Can you make it so that the context menu only appears when the mouse is over an actual entry?

I'll see if I can make it work like that. Otherwise, I'll try to add the right-aligned 'X' close button.

cbrnr:
@GuillaumeFavelier do you want to continue working on this? We can skip the X button and only have a context menu, but it should only appear when right-clicking on an item. The menu could include two entries "Rename" and "Close" for now.

cbrnr:
Superseded by #454.

Issue #312, opened 2022-02-19
Use infowidget when no file is open


cbrnr:
I think we could use the infowidget to display information like which file formats can be opened, shortcuts, etc. instead of just an empty gray area when no dataset is open. Inspired by VS Code, which I really like:
<img width="930" alt="Screen Shot 2022-02-19 at 11 56 51" src="https://user-images.githubusercontent.com/4377312/154798075-37675eab-4376-4444-b9aa-16e368c619b2.png">

GuillaumeFavelier:
Do you also consider the toolbar on the left? :grin:

cbrnr:
No, here I just meant the white empty editor space 😄!

PR #343, opened 2022-03-31
WIP,ENH: info widget shortcuts

GuillaumeFavelier:
This PR populates the `InfoWidget` at init with the default shortcuts:

image

For now it reuses `set_values()` because it's the most convenient but to achieve something closer to the desired look of #312, something more elaborate is required.

Closes https://github.com/cbrnr/mnelab/issues/312

### DIFF
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -360,7 +360,7 @@ class MainWindow(QMainWindow):
 
         self.setAcceptDrops(True)
         self.data_changed()
-        self.infowidget.set_values(shortcuts)
+        self.infowidget.set_values(shortcuts, shortcut=True)
 
     def _excepthook(self, type, value, traceback_):
         exception_text = str(value)
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -1331,4 +1331,8 @@ class MainWindow(QMainWindow):
                 QIcon.setThemeName(style)
             else:
                 QIcon.setThemeName("light")  # fallback
+            # update shortcuts stylesheet if necessary
+            if hasattr(self, "infowidget"):
+                if self.infowidget._shortcut:
+                    self.infowidget.set_shortcuts()
         return super().event(ev)
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -277,7 +277,7 @@ class MainWindow(QMainWindow):
 
         view_menu = self.menuBar().addMenu("&View")
         self.actions["history"] = view_menu.addAction("&History", self.show_history,
-                                                      QKeySequence(Qt.CTRL + Qt.Key_H))
+                                                      QKeySequence(Qt.CTRL + Qt.Key_Y))
         self.actions["toolbar"] = view_menu.addAction("&Toolbar", self._toggle_toolbar)
         self.actions["toolbar"].setCheckable(True)
         self.actions["statusbar"] = view_menu.addAction("&Statusbar",
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -276,7 +276,8 @@ class MainWindow(QMainWindow):
         )
 
         view_menu = self.menuBar().addMenu("&View")
-        self.actions["history"] = view_menu.addAction("&History", self.show_history)
+        self.actions["history"] = view_menu.addAction("&History", self.show_history,
+                                                      QKeySequence(Qt.CTRL + Qt.Key_H))
         self.actions["toolbar"] = view_menu.addAction("&Toolbar", self._toggle_toolbar)
         self.actions["toolbar"].setCheckable(True)
         self.actions["statusbar"] = view_menu.addAction("&Statusbar",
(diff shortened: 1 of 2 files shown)


GuillaumeFavelier:
With rough modifications of the style:

image

cbrnr:
Looks great! A couple of suggestions:

- Can you center it inside the info widget (both horizontally and vertically)?
- macOS uses ⌘ (cmd) instead of Ctrl – maybe there's a way to get this from Qt directly? Otherwise there should be a condition for macOS.
- Can you adapt the colors to the current theme (light/dark)? This only works on macOS, but it would look super cool if the buttons were also dark.
- I'd prefer the buttons to be a bit smaller.
- I think we should add a shortcut for View - History (e.g. ⌘ + H).
- WDYT, should we maybe manually select the shortcuts we want to show? There are not many active anyway, and ⌘ + W (Close) doesn't make sense because nothing is open.

Screen Shot 2022-03-31 at 20 03 49

GuillaumeFavelier:
After the latest commits:

light | dark
--|--
image | image

A few things to do:
* add the real 'Ctrl + H' shortcut
* support dynamic theme?
* load shortcuts automatically when the last data is closed?

cbrnr:
> add the real 'Ctrl + H' shortcut

This should be easy.

> support dynamic theme?

Yes! Maybe move the stylesheet to `utils` where `interface_style()` is also defined?

> load shortcuts automatically when the last data is closed?

Yes, this was the idea. Instead of showing a blank area, the shortcuts should be shown.

BTW since when does the dark theme work? I have only ever seen this working on macOS. Does it work dynamically?

GuillaumeFavelier:
> Does it work dynamically?

Ah sorry, the advantage of using the dev version, I just changed the default for the screenshot  :sweat_smile: . But an idea to make the life of linux/windows users easier would maybe to use an environment variable `MNELAB_THEME` or something.

GuillaumeFavelier:
I pushed d45185021aae8fbe141a891ba23956209ac2afd5 to add dynamic support (in theory). Feel free to try it out @cbrnr.

cbrnr:
Yes, it works dynamically!

However, ⌘ + H is a reserved system shortcut on macOS, so could you change it to ⌘ + Y? This could be used on all platforms for simplicity.

So you are saying that dynamic theme switching does *not* work in KDE, right?

But I'm still surprised that the dark theme works at all, because last time I tried this on Linux it didn't work. However, I'm using GNOME, so maybe some library/package is missing? Do you know what I could be missing?

Final nitpick: would it be possible to arrange the shortcut in an invisible table with two columns, so that text in the left column is right-aligned and the keys in the left column are left aligned?

cbrnr:
And yes, good idea to add some option to manually set the theme. We already have a preferences dialog, which would be perfect for this purpose (a combobox with "light", "dark", and "auto").

cbrnr:
> It works for the default apps but it is not complete for mnelab, I have to restart the app:

I see, but at least it does work (in contrast to Windows). Apparently KDE emits a different signal that Qt must be able to catch, so in principle we could try to make it work. But this is something for another PR I guess.

> It's a long time since I used Gnome, I cannot really tell but AFAIK it should be possible to apply a global dark theme in the same way 😅

It's not important. As long as there is no desktop-wide standard it will always be difficult to make it work for multiple environments.

> I can look into it but it might take a while, this is where QtDesigner really shines usually

Yes, I know – I avoided it because there are only simple dialogs so far. Maybe hacking it together in Designer and then looking at the generated code will make it easier? If it's too much effort don't bother and we can always work on that later.

cbrnr:
This separate widget could be a rich text widget, e.g.

```python
from PySide6.QtWidgets import QApplication, QLabel, QWidget, QVBoxLayout


def make_shortcuts_table(names, shortcuts, style="light"):
    text_color = "#777"
    if style == "light":
        background_color = "#444"
    else:
        background_color = "#444"
    html = f"""<!DOCTYPE html>
    <html>
      <head>
        <style>
          html {{ font-size: 16px; }}
          kbd {{
            background-color: {background_color};
            font-weight: 600;
            color: {text_color};
          }}
          table {{ color: {text_color}; }}
        </style>
      </head>
      <body>
        <table>
          <tbody>"""
    for name, shortcut in zip(names, shortcuts):
        modifier, key = shortcut
        html += (
            f'\n            <tr><td align="right" width="50%">{name} </td>'
            f'<td><kbd>{modifier}</kbd> '
            f'<kbd>{key}</kbd></td></tr>'
        )
    html += """\n          </tbody>
        </table>
      </body>
    </html>"""
    return html


app = QApplication()
window = QWidget()
table = make_shortcuts_table(["Open", "Close", "History"], ["⌘O", "⌘W", "⌘Y"], style="x")
print(table)
text = QLabel(table)
vbox = QVBoxLayout()
vbox.addStretch()
vbox.addWidget(text)
vbox.addStretch()
window.setLayout(vbox)
window.show()
app.exec()
```

But your buttons look a bit better I think.

cbrnr:
Superseded by #350.

PR #363, opened 2022-10-18
Limit PySide6 minor version to 6.3

### LINKED ISSUE #362: Runtime breaks with pyside6 v6.4.0
Installing and running MNELAB in a new environment raises `TypeError: int() argument must be a string, a bytes-like object or a number, not 'KeyboardModifier'` (see matplotlib/matplotlib#24155 for more information).
I think a reasonable solution until this is resolved in pyside is to limit the required version for `mnelab`.

ZviBaratz:
Resolves #362.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,7 +22,7 @@ packages = find:
 python_requires = >=3.8
 install_requires =
     mne>=1.0.0
-    PySide6>=6.2.0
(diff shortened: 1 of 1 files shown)


cbrnr:
Matplotlib 3.6.2 fixed the issue.

PR #372, opened 2023-04-21
Migrate setuptools to pyproject.toml

cbrnr:
According to https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html, setting up the build toolchain should now work with `pyproject.toml`.


cbrnr:
@larsoner you have a lot of experience with issues related to running tests on CI servers – do you happen to know what's going on here (https://github.com/cbrnr/mnelab/actions/runs/4766432481/jobs/8473493366?pr=372)?

larsoner:
It's very likely you are missing some dependencies of the Qt libraries. To stay up to date for free I'd use pyvista/setup-headless-display-action with `qt: true` as in:

https://github.com/pyvista/pyvistaqt/blob/c1f45e98b7c9b6512559e4927a5a175a8da8d6b3/.github/workflows/ci.yml#L31-L34

To check you can try just creating a QApplication separately like:

https://github.com/mne-tools/mne-python/blob/4eb9ef04c50710dba840ceaac6dc4f69ffefdd31/azure-pipelines.yml#L131

Every few months it seems there is some new dependency the PyQt folks decide to no longer bundle and require at the system level. I plan to keep the `setup-headless-display-action` up to date with these requirements.

cbrnr:
Thanks @larsoner, I'll try that action! This would be a relief!

cbrnr:
Unfortunately, this doesn't work either 😢.

larsoner:
> To check you can try just creating a QApplication separately like:

Did you try this suggestion?

cbrnr:
> Did you try this suggestion?

Oops, sorry, missed that. Will try!

cbrnr:
Not really sure what's going on. @larsoner do you know?

.github/workflows/test.yml
```diff
@@ -49,13 +49,14 @@ jobs:
...
-    - name: Install Linux packages
+    - name: Debug
       run: |
-        sudo apt update
-        sudo apt install libxcb-icccm4 libxcb-image0 libxcb-keysyms1 \
-                         libxkbcommon-x11-0 xvfb libxcb-randr0 \
-                         libxcb-render-util0 libxcb-xinerama0 libegl1 \
-                         libxcb-shape0
+        set -e
+        LD_DEBUG=libs python -c "from PySide6.QtWidgets import QApplication, QWidget; app = QApplication([]); import matplotlib; matplotlib.use('QtAgg'); import matplotlib.pyplot as plt; plt.figure()"
```
larsoner, line 55: The error log says:
```
ModuleNotFoundError: No module named 'PySide6'
```
This step needs to go after PySide6 has been installed, which would be after the "Install Python Packages" step

cbrnr: Got it. New error: https://github.com/cbrnr/mnelab/actions/runs/4768501192/jobs/8477907996?pr=372#step:7:852

.github/workflows/test.yml
```diff
@@ -49,16 +49,17 @@ jobs:
...
       with:
         name: mnelab-wheel
-    - name: Install Linux packages
-      run: |
-        sudo apt update
-        sudo apt install libxcb-icccm4 libxcb-image0 libxcb-keysyms1 \
-                         libxkbcommon-x11-0 xvfb libxcb-randr0 \
-                         libxcb-render-util0 libxcb-xinerama0 libegl1 \
-                         libxcb-shape0
+    - uses: pyvista/setup-headless-display-action@v2
```
larsoner, line 52: This fix is in `main` but not released
```suggestion
    - uses: pyvista/setup-headless-display-action@main
```

cbrnr:
I decided to just install the missing package for now, but I might come back to your action in the future. I just don't want to use the main branch of an action, this seems a bit too cutting edge. Anyway, thanks for the help @larsoner!

PR #379, opened 2023-05-10
Set build-system.build-backend in pyproject.toml

musicinmybrain:
See: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html


cbrnr:
Thanks @musicinmybrain!

PR #382, opened 2023-05-29
fix ERDS plot with mne >= 1.1.0, fix #381

### LINKED ISSUE #381: ERDS plots not working, fix needed with newer MNE
Hello,

I am following the tutorial to plot ERDS, but the dialog "Calculating ERDS maps..." takes forever... and never ends. This happens on W10 with 0.8.6 as well as un Ubuntu 20.04 with 0.8.5 and 0.8.0 (what I could test with my older python there). After some debugging I found out that there is an error upon execution of the underlying MNE command (an error hidden by the use of the dialog in the GUI). `baseline` appears to now be a tuple, and not a list anymore, at least with mne 1.4.0 that is installed on both my systems. If I change `baseline=[dialog.b1, dialog.b2]` to `baseline=(dialog.b1, dialog.b2)` in `plot_erds` now everything works fine. I also changed it in `plot_erds_topomaps`, even though I did not test this one.

The change is trivial but I guess you might need some workarounds to handle different MNE versions, and possibly this change is necessary in other parts of the tool -- which is very handy to get to know MNE :)

Hope it helps!

Edit: the change in MNE might be circa https://github.com/mne-tools/mne-python/commit/64dff6189215cbead61db81d6e6de811a05633e3

jfrey-xx:
Here is the tiny change so that ERDS plot can work with newer MNE, as discussed in #381.

### DIFF
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -875,7 +875,7 @@ class MainWindow(QMainWindow):
 
         if dialog.exec():
             freqs = np.arange(dialog.f1, dialog.f2, dialog.step)
-            baseline = [dialog.b1, dialog.b2]
+            baseline = (dialog.b1, dialog.b2)
             times = [dialog.t1, dialog.t2]
             alpha = None
             if dialog.significance_mask.isChecked():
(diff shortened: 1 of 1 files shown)


cbrnr:
Thanks a lot! I'm not sure if this change was made on purpose, so I've asked in https://github.com/mne-tools/mne-python/issues/11711. But in any case, using a tuple is a good choice here.

jfrey-xx:
Thanks for the quick reaction! I did not think about the changelog, I'll keep that in mind if I am to change something else :)

PR #403, opened 2024-01-21
FIX: Work with Qt backend

marsipu:
This makes the raw-plot from inside mnelab work with the qt-backend.


marsipu:
At the moment, the performance of the qt-backend takes a huge hit when opened from inside mnelab (not only because it plots all channels). I couldn't figure out why yet. I first thought it was because in mainwindow there is the method `event` catching all events but it doesn't seem to be it. But I suspect something similar, where Qt-Backends are caught and altered in python-code. @cbrnr Do you maybe have an idea?
When I start a plot from inside my GUI in [mne-pipeline-hd](https://github.com/marsipu/mne-pipeline-hd), there is no performance hit. And I am catching bad channels the same way as you do.

cbrnr:
Unfortunately, it still crashes on macOS 😢, so I cannot even test performance. I have no idea why it is so slow, I'm not catching any other events.

cbrnr:
BTW, I tried mne-pipeline-hd, and it crashes with the same error (after creating the project folder).

cbrnr:
I think the safest thing for now is to force the Matplotlib backend, given that the qt backend (1) crashes on macOS and (2) is very slow. Once we figure out these two issues, I'm happy to re-enable it.

cbrnr:
Do you want to implement this @marsipu (keeping the new event handling for bad channels)? If not, I'm happy to take over.

marsipu:
@cbrnr thanks for unearthing this macos-qt-problem, hopefully we find a solution in the future.
I'd be glad if you took over, since I am quite busy with startig my new job at the hospital at the moment 🏥

cbrnr:
Sure, thanks @marsipu and all the best for your new job!

danieltomasz:
Pinning  PySide6 to 6.5.3 and reverting to 0.8.6 fixes crashing on MacOS Sonoma

cbrnr:
Is reverting to 0.8.6 necessary?

danieltomasz:
@cbrnr maybe not,  I downgrade because `mnelab 9.0`  has pinned   `"PySide6 >= 6.6.0"`, 
unfortunately  there is  a caveat - `PySide6 6.5.3` doesn't support `Python 3.12`

cbrnr:
> unfortunately there is a caveat - PySide6 6.5.3 doesn't support Python 3.12

This is really too bad! Now I wonder what to do. I don't want to drop Python 3.12 support, and it will probably take a while until this bug is fixed in PySide6 (Qt?). I guess we'll have to stick with the Matplotlib-based browser until this issue has been resolved...

danieltomasz:
@cbrnr `PyQt v6.5.3` released in October seems to support `Python 3.12` and python apps using this version seems to be able to fix this bug; 
https://github.com/jim-easterbrook/Photini/issues/229
If you try to pin PyQt depenedcy this fix  this migth work, the error  in the comment above https://github.com/cbrnr/mnelab/pull/403#issuecomment-1903699069   is showing that you have installed into your env `PyQt v6.6.1` which have the same problem  as   `PySide6 6.6`

cbrnr:
OK, I just found out that `PySide6==6.6.0` works (only later versions like 6.6.1 are affected by the crash). So that's at least something. However, when the Qt Browser window opens, the icons in MNELAB disappear and contain only text...

Screenshot 2024-02-12 at 08 20 19

cbrnr:
Here's the Qt issue again, hopefully this is the same issue and they can fix the problem: https://bugreports.qt.io/browse/QTBUG-120181

cbrnr (commented):
@marsipu I'm trying to disable the <kbd>Esc</kbd> key for closing the figure. The following solution works for the `matplotlib` backend:

```
fig.mne.close_key = None
```

However, this has no effect for the `qt` backend. Do you know how this can be achieved?

cbrnr (commented):
Thanks @marsipu!

PR #415, opened 2024-03-15
Add browser backend setting

cbrnr:
This PR adds a setting to select the browser backend (default "Matplotlib", and if mne-qt-browser is installed optionally "qt").

### DIFF
--- a/mnelab/mainwindow.py
+++ b/mnelab/mainwindow.py
@@ -74,15 +74,21 @@ class MainWindow(QMainWindow):
         # remove None entries from self.recent
         self.recent = [recent for recent in self.recent if recent is not None]
 
+        # plot backend
+        self.plot_backends = ["Matplotlib"]
+        if have["mne-qt-browser"]:
+            self.plot_backends.append("Qt")
+        plot_backend = settings["plot_backend"]
+        if plot_backend not in self.plot_backends:
+            plot_backend = "Matplotlib"
+        mne.viz.set_browser_backend(plot_backend)
+
         # trigger theme setting
         QIcon.setThemeSearchPaths(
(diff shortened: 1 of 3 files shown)


cbrnr (commented):
Note that this PR requires a new release for mne-qt-browser (I've assumed it will be 0.6.2) with all changes related to icon themes and bundle name.

cbrnr:
Ouch. When setting the plot backend to `"qt"`, I get a segfault on Linux (Wayland) before the main window even opens. I guess this is caused by `mne.viz.set_browser_backend("qt")` during `__init__`:

```
[1]    33942 segmentation fault (core dumped)  mnelab
```

:cry:

cbrnr:
An even simpler MWE (uses `from mne_qt_browser import _pg_figure` and does not need `import mne` anymore, because that's what's happening when doing `mne.viz.set_browser_backend("qt")`):

```python
import sys

from PyQt6.QtWidgets import QApplication, QMainWindow
from mne_qt_browser import _pg_figure


app = QApplication(sys.argv)
app_name = "qt-browser-backend-issue"
app.setApplicationName(app_name)

window = QMainWindow()
window.setWindowTitle(app_name)
window.show()

sys.exit(app.exec())
```

cbrnr:
Related upstream issue: https://bugreports.qt.io/browse/PYSIDE-2676

larsoner:
See also https://bugreports.qt.io/browse/PYSIDE-2674 which could be related

To make things easier for the PySide folks I'd suggest working to get rid of the `mne-qt-browser` requirement, it could for example be due to `pyqtgraph` which you could then whittle down etc. Could be due to PyOpenGL being installed or not, some qtpy bug, etc. and without trimming down the example it's tough to know

PR #417, opened 2024-04-09
Use built-in MNE export for BrainVision

cbrnr:
This outsources BrainVision export to built-in MNE export functionality. Eventually, this should make it possible to export arbitrary markers (without converting them to `S  1` etc.).


cbrnr:
There's an additional marker containing the measurement date in the exported file, which should probably be removed on import (see https://github.com/mne-tools/mne-python/issues/13095).

PR #419, opened 2024-04-11
Switch build backend to hatchling

cbrnr:
- [x] Switch build backend to hatchling
- [x] Switch to src layout
- [x] Use `python -m build` as a build frontend in workflows
- [x] Use trusted publishing on PyPI


pyproject.toml
```diff
@@ -54,6 +57,9 @@ documentation = "https://mnelab.readthedocs.io/en/latest/"
 repository = "https://github.com/cbrnr/mnelab"
 changelog = "https://github.com/cbrnr/mnelab/blob/main/CHANGELOG.md"
 
+[tool.hatch.version]
+path = "src/mnelab/__init__.py"
```
hoechenberger, line 61: why is this needed?

cbrnr: Because that's where the version is defined. This will become obsolete when I switch to `hatch-vcs` though.

cbrnr (commented):
Merging, let's tackle `hatch-vcs` in another PR.

hoechenberger:
@cbrnr I believe you forgot to remove `MANIFEST.in`

PR #434, opened 2024-09-17
Refactor readers to subclass `BaseRaw`

cbrnr:
This is the recommended and documented approach, which specifically means that we do not have to fiddle with the private `._filenames` attribute anymore.

- [x] `.mat`
- [x] `.npy`
- [x] `.xdf`/`.xdfz`

This is what you meant, right @larsoner?


larsoner:
Looks reasonable to me

PR #446, opened 2024-10-29
Show correct history for appending data

### LINKED ISSUE #130: Show correct history for appending data
Currently, the history for appending data does not reflect the correct MNE command (it shows data set names instead of data set objects, i.e. we need to show something like `[datasets[1], datasets[3]]`, but before this works #127 needs to be merged).

bkloeckl:
Fixes #130. Changes in model.py: def append_data(self, names) to correctly show the code history

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 ## [UNRELEASED] - YYYY-MM-DD
 ### Fixed 
-- Fixed a bug where appending data would not be correctly displayed in the history ([#446]) https://github.com/cbrnr/mnelab/pull/446 by [bkloeckl] (https://github.com/bkloeckl))
+- Fix a bug where appending data would not be correctly displayed in the history ([#446]) https://github.com/cbrnr/mnelab/pull/446 by [Benedikt Kloeckl](https://github.com/bkloeckl))
+
+### Added
+- Add tests for model.append_data() by generating temporary .edf files by [Benedikt Kloeckl](https://github.com/bkloeckl)).
 
 ## [0.9.2] - 2024-10-20
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## [UNRELEASED] - YYYY-MM-DD
+### Fixed 
+- Fixed a bug where appending data would not be correctly displayed in the history ([#446]) https://github.com/cbrnr/mnelab/pull/446 by [bkloeckl] (https://github.com/bkloeckl))
 
 ## [0.9.2] - 2024-10-20
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,6 @@
 ### Fixed 
 - Fix a bug where appending data would not be correctly displayed in the history ([#446]) https://github.com/cbrnr/mnelab/pull/446 by [Benedikt Kloeckl](https://github.com/bkloeckl))
 
-### Added
-- Add tests for model.append_data() by generating temporary .edf files by [Benedikt Kloeckl](https://github.com/bkloeckl)).
-
 ## [0.9.2] - 2024-10-20
 ### Added
 - Add initial number of displayed channels in the raw plot to settings dialog and set default to 32 channels ([#428](https://github.com/cbrnr/mnelab/pull/428) by [Dennis Wambacher](https://github.com/d3njo))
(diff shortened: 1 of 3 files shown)


cbrnr:
I think duplicate names might still be a problem as you mentioned in #130. In my example with two identical names, if you only select one S001R04 in the append dialog, MNELAB still appends both datasets.

So I think we need a different solution. In general, there are two options:

1. Disallow duplicate names (which you suggested)
2. Use the underlying indices in functions that operate on multiple datasets

The first option is probably easier to implement. The second option would require a couple of changes, e.g. showing the indices in the sidebar and all dialogs that require multiple datasets. WDYT?

cbrnr:
Let's continue the discussion in #446!

src/mnelab/model.py
```diff
@@ -533,19 +533,24 @@ def get_compatibles(self):
...
+        files = []
+        indices = []
+
+        for idx, d in enumerate(self.data):
             if d["name"] in names:
                 files.append(d["data"])
+                indices.append(f"datasets[{idx}]")
+            elif d["name"] == self.current["name"]:
+                files.insert(0, d["data"])
+                indices.insert(0, f"datasets[{idx}]")
```
cbrnr, line 546: I think it would be easier if you first insert the current dataset, and then in the loop insert the to-be-appended datasets (so everything except for the first). There's not need for the if/elif in the loop.

bkloeckl: That would work for files[], but the index of the current entry can only be identified correctly in the loop, as it is dependent on whether or not the dataset was duplicated or will be overwritten.

cbrnr: But the current dataset is also available as `data`, so the resulting history command is always `mne.concatenate_raws([data, ...])`, or no?

bkloeckl: That could also work. I will look into it, if it that would result in a cleaner solution.

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 ## [UNRELEASED] - YYYY-MM-DD
 ### Fixed 
-- Fixed a bug where appending data would not be correctly displayed in the history ([#446]) https://github.com/cbrnr/mnelab/pull/446 by [bkloeckl] (https://github.com/bkloeckl))
+- Fix a bug where appending data would not be correctly displayed in the history ([#446]) https://github.com/cbrnr/mnelab/pull/446 by [Benedikt Kloeckl](https://github.com/bkloeckl))
+
+### Added
+- Add tests for model.append_data() by generating temporary .edf files by [Benedikt Kloeckl](https://github.com/bkloeckl)).
```
cbrnr, line 6: This is a user-facing document, so I would not mention any changes to tests here.

cbrnr:
I don't know why monkeypatching the decorator doesn't work, but at least this is a workaround. To improve upon that a little, I think that we could think about adding two new parameters `view=None` and `decorator=None` to the `Model` initializer, which would probably make it a bit easier to use a model without a view. Or maybe we only need `view=None` and we adapt the decorator so that it can handle cases where `args[0].view = None`. This would probably be the nicest solution so far.

src/mnelab/model.py
```diff
@@ -43,8 +43,8 @@ def wrapper(*args, **kwargs):
 class Model:
     """Data model for MNELAB."""
 
-    def __init__(self):
-        self.view = None  # current view
+    def __init__(self, view=None):
+        self.view = view  # current view, default = None
```
cbrnr, line 47: I actually just noticed that this does not work for initializing our model in one step, because the view requires a model, and so the model cannot require a view. So maybe let's leave it as is, meaning that by default `model.view = None`, there is no specific parameter needed. Sorry about the confusion!

tests/test_model.py
```diff
@@ -0,0 +1,157 @@
...
+    class DummyView:
+        @contextmanager
+        def _wait_cursor(self):
+            yield
+
+        def data_changed(self):
+            pass
+
+    model_instance = Model(view=DummyView())
+    return model_instance
```
cbrnr, line 31: Here I would prefer to simply adapt the `data_changed` decorator so that it can handle cases where `view is None`.

cbrnr:
I have parameterized the test to avoid repeating code, and I've also used the `:=` operator to shorten some things. I think it looks really nice now. Thanks a lot for your PR @bkloeckl 🚀!

PR #449, opened 2024-11-17
Use unique indices to identify datasets for appending

bkloeckl:
Change the append dialog, get_compatibles and append_data functions to include original and unique indices used for identifying the data.


cbrnr:
Thanks @bkloeckl! The current changes lack the ability to sort datasets (via drag and drop) in the Source and Destination lists though, which I think we should not remove.

I would also keep it simpler, instead of using a table, I'd stick to a list, but just add the index to the label. This could also be done for the sidebar then. I'm not sure how and if such a label can be formatted nicely, e.g. it would be nice to have

- numbers right-aligned
- numbers use some lighter color (not black or white)

Something like this:

Screenshot 2024-11-18 at 09 17 41

If this requires a table (which might very well be the case), I'm sure that the style can be adapted so that it doesn't contain a header row or any borders.

src/mnelab/dialogs/append.py
```diff
@@ -90,16 +90,23 @@ def dropEvent(self, event):
                     item.setTextAlignment(
                         Qt.AlignLeft | Qt.AlignVCenter
                         if col == 1
-                        else Qt.AlignRight | Qt.AlignVCenter
+                        else Qt.AlignVCenter | Qt.AlignVCenter
```
cbrnr, line 93: Typo?

bkloeckl: Yes, remnants of the styling iterations. Just corrected

cbrnr:
I found an issue with the current implementation:

https://github.com/user-attachments/assets/20c43ee6-9810-4ea0-b23b-4c59116d1504

cbrnr:
Another thing that could be improved is that currently, dragging one or more items from the source table to the destination table (or vice versa) shows the drop hint (the black line where the item would be inserted), but unfortunately dropping to the other table does not work. I think it would be very nice if that worked (then users could operate the dialog only via drag and drop and would not even need the arrow button, which is still a useful backup/alternative).

cbrnr:
Finally, I was wondering if you really have to implement the drag and drop on that level, or if you could maybe just connect the `rowsMoved` signal to simplify (as is done [here](https://github.com/bkloeckl/mnelab/blob/main/src/mnelab/mainwindow.py#L383) for drag and drop in the sidebar)?

bkloeckl:
> Finally, I was wondering if you really have to implement the drag and drop on that level, or if you could maybe just connect the `rowsMoved` signal to simplify (as is done [here](https://github.com/bkloeckl/mnelab/blob/main/src/mnelab/mainwindow.py#L383) for drag and drop in the sidebar)?

For the sidebar, the data and its order are managed by the model and then visually updated. The append-dialog has to handle the reordering on its own. This implementation is also the base for the latest update, where drag&drop works across the source and destination tables.

cbrnr:
Thanks @bkloeckl!

PR #454, opened 2024-11-26
Sidebar overhaul

bkloeckl:
Change sidebar UIWidget from List to Table in order to display unique indices of datasets and add a close-button in each row.


cbrnr:
I have two suggestions regarding the behavior of this implementation:

1. The "x" button should disappear when the mouse is not hovering over any data set. Currently, the "x" stays where it last was when the mouse leaves the widget, which I find distracting.
2. Closing a dataset is not yet reflected in the history. I know that this was also the case before this PR, but maybe you could add a history entry for the close data method as part of this PR?

I haven't looked in detail at the changes (will do that soon), but because you changed `model.remove_data()`, I was wondering if things like `self.model.index -= 1` after `self.model.remove_data()` still work as intended (see [here](https://github.com/cbrnr/mnelab/blob/ada7a1659f6819c73776c886a7074fa66ac9fa42/src/mnelab/mainwindow.py#L1099-L1100))? There's even a line that is commented out (see [here](https://github.com/cbrnr/mnelab/blob/ada7a1659f6819c73776c886a7074fa66ac9fa42/src/mnelab/mainwindow.py#L1235)). In total, there are three occurrences of this pattern (in `interpolate_bads()`, `epoch_data()`, and `change_reference()`). Can you check what's up with those?

Also, as we've already discussed, I think you do not have to duplicate the drag and drop behavior (for dropping files), since this is already implemented in `MainWindow`.

I'll review the changes in detail after we've addressed these points.

bkloeckl:
> I haven't looked in detail at the changes (will do that soon), but because you changed `model.remove_data()`, I was wondering if things like `self.model.index -= 1` after `self.model.remove_data()` still work as intended (see [here](https://github.com/cbrnr/mnelab/blob/ada7a1659f6819c73776c886a7074fa66ac9fa42/src/mnelab/mainwindow.py#L1099-L1100))? There's even a line that is commented out (see [here](https://github.com/cbrnr/mnelab/blob/ada7a1659f6819c73776c886a7074fa66ac9fa42/src/mnelab/mainwindow.py#L1235)). In total, there are three occurrences of this pattern (in `interpolate_bads()`, `epoch_data()`, and `change_reference()`). Can you check what's up with those?
> 

The changes in the `remove_data()` function now allow to pass an index parameter, necessary for deleting rows, not currently selected. Its functionality, when no index is passed, remains however unchanged.

cbrnr:
This is a very nice change, I like it a lot! Thanks @bkloeckl!

PR #457, opened 2024-12-06
Add tests on Python 3.13

cbrnr:
Since MNELAB has been compatible with 3.13 for some time, I think it was an oversight that our tests didn't include that version.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,7 @@ jobs:
     needs: style
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


cbrnr (commented):
Of course, it probably wasn't an oversight. Running into the `uv` resolution issue again (it tries to resolve fixed package versions across *all* supported Python versions, which means that since 3.9 only supports `scipy==1.13.1` and not the latest `1.14.1`, it picks `1.13.1` but then has to compile from source, because `1.13.1` is not available as a binary wheel for Python 3.13).

Issue #253, opened 2022-02-02
Montage previews cannot be closed on macOS


cbrnr:
When previewing a montage (_Edit_ – _Set montage..._, select any montage and click on _View_), the montage previews that open in a separate window cannot be closed as long as the _Set montage_ dialog is still open.

hofaflo:
Cannot reproduce on Win10, multiple montage preview windows can be closed independent of the "Set montag" dialog's state 🤔

cbrnr:
A nice solution would be to include the previews in the montage dialog, perhaps with a collapsible "Details" button.

PR #459, opened 2024-12-06
Change Set Montage dialog view

bkloeckl:
To resolve [Issue #253](https://github.com/cbrnr/mnelab/issues/253) while also enhancing the user experience, the preview of the selected channel montage should be embedded within the same window, rather than opening in a new window.


cbrnr:
Nice! I have some cosmetic suggestions:

Screenshot 2024-12-06 at 14 04 15

Also, I think there should be a minimal width of the preview plot.

cbrnr:
> However I would not place the buttons on the bottom-right corner as they would appear "lost" there. In my opinion they fit way better to the other control elements also for usability purposes. Let me know, what you think about the updated version.

Yes, but I actually have plans for this bottom row, because I would like to add additional buttons in the future (e.g., "Read custom montage file..."). Can you show how this would look like now (with only the OK/Cancel buttons in the bottom row) so that we can have a visual comparison?

cbrnr:
Thanks @bkloeckl!

Issue #340, opened 2022-03-31
Show channel stats


cbrnr:
I think it would be useful to add a new option to show channel stats as obtained with `raw.describe()` (which returns a Pandas dataframe), e.g.:

```ch  name  type  unit        min         Q1     median         Q3        max
 0  Fp1   EEG   µV      -321.63     -13.21      -1.94       8.68     712.39
 1  AF7   EEG   µV      -320.12      -9.66      -0.20       9.44     356.74
 2  AF3   EEG   µV      -282.45      -7.04       0.12       7.40     234.54
 3  F1    EEG   µV      -389.73     -21.54      -2.57      16.46     708.45
 4  F3    EEG   µV      -334.93     -13.89      -2.23       9.02     378.99
 5  F5    EEG   µV      -458.99      -7.28      -0.47       6.42     300.37
 6  F7    EEG   µV      -315.50      -7.62      -0.24       7.49     236.63
 7  FT7   EEG   µV     -1200.60      -7.71       0.35       8.36    2662.98
 8  FC5   EEG   µV      -422.04      -6.93       0.32       7.46     232.00
 9  FC3   EEG   µV      -253.39     -11.36      -0.04      11.36     405.27
10  FC1   EEG   µV      -249.88      -6.05       0.46       6.85     117.05
11  C1    EEG   µV      -236.54     -10.62      -0.19      10.44     133.33
12  C3    EEG   µV      -282.18      -5.73       0.22       6.08     122.19
13  C5    EEG   µV      -325.40      -8.38      -0.18       8.09     303.00
14  T7    EEG   µV      -547.49     -25.18      -1.26      23.27     949.80
15  TP7   EEG   µV      -315.58     -15.28      -1.11      13.25     430.67
16  CP5   EEG   µV      -246.36      -6.91      -0.12       6.80     133.41
17  CP3   EEG   µV      -185.79      -9.15      -0.31       8.74     229.63
18  CP1   EEG   µV      -201.36      -6.47       0.33       6.75      85.55
19  P1    EEG   µV      -263.86     -11.62       0.28      12.20     178.38
20  P3    EEG   µV      -191.56      -5.75       0.06       5.85      77.31
21  P5    EEG   µV      -208.69      -5.75       0.11       5.83     116.09
22  P7    EEG   µV      -217.80      -6.66      -0.14       6.52     219.52
23  P9    EEG   µV      -274.54      -8.76      -0.09       8.76     366.36
[...]

PR #462, opened 2024-12-12
Show channel stats

bkloeckl:
Add new view dialog to display channel stats from raw.describe() in a table. Issue #340

### DIFF
--- a/src/mnelab/dialogs/channel_stats.py
+++ b/src/mnelab/dialogs/channel_stats.py
@@ -1,0 +1,46 @@
+# © MNELAB developers
+#
+# License: BSD (3-clause)
+from PySide6.QtWidgets import (
+    QDialog,
+    QPushButton,
+    QTableWidget,
+    QTableWidgetItem,
+    QVBoxLayout,
+)
+
+
+class ChannelStats(QDialog):
+    def __init__(self, parent, data):
+        super().__init__(parent=parent)
+
+        # window
+        self.setWindowTitle("Channel Stats")
+        self.resize(690, 800)
+        self.setMinimumSize(200, 150)
+        self.setMaximumWidth(690)
+        layout = QVBoxLayout(self)
+
+        # table widget
+        self.table = QTableWidget()
+        self.table.setRowCount(len(data))
+        self.table.setColumnCount(len(data.columns))
+        self.table.setHorizontalHeaderLabels(data.columns)
+
+        # populate table
+        for row in range(len(data)):
+            for col in range(len(data.columns)):
+                item = QTableWidgetItem(str(data.iloc[row, col]))
+                self.table.setItem(row, col, item)
+
+        self.table.resizeColumnToContents(0)  # name
+        self.table.resizeColumnToContents(1)  # type
+        self.table.resizeColumnToContents(2)  # unit
+        layout.addWidget(self.table)
+
+        # add close button
+        close_button = QPushButton("Close")
+        close_button.clicked.connect(self.close)
+        layout.addWidget(close_button)
+
+        self.setLayout(layout)
--- a/src/mnelab/mainwindow.py
+++ b/src/mnelab/mainwindow.py
@@ -29,6 +29,7 @@ from PySide6.QtWidgets import (
 from pyxdf import resolve_streams
 
 from mnelab.dialogs import *  # noqa: F403
+from mnelab.dialogs.channel_stats import ChannelStats
 from mnelab.io import writers
 from mnelab.io.mat import parse_mat
 from mnelab.io.npy import parse_npy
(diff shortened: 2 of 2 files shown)


cbrnr:
Nice, just some additional thoughts:

- I think you can remove the dependency on pandas.
- It would be great if the calculation was available somewhere outside the dialog, maybe somewhere in the `utils` subpackage? This would decouple the calculation from the dialog.
- The -1 unit has quite a long description (FIFF_UNIT_NONE), it would be good to maybe just say "None" instead (e.g. for STIM channels).
- Numeric columns should be right-aligned, and all values should have the same number of decimal places (e.g. 1 or 2).
- I think a mean column would also be useful (maybe just before the median column).

pyproject.toml
```diff
@@ -22,16 +22,17 @@ classifiers = [
...
 dependencies = [
-    "mne >= 1.7.0",
-    "PySide6 >= 6.7.1",
     "edfio >= 0.4.2",
     "matplotlib >= 3.8.0",
+    "mne >= 1.7.0",
     "numpy >= 2.0.0",
-    "scipy >= 1.14.1",
     "pybv >= 0.7.4",
+    "pyobjc-framework-cocoa >= 10.0; platform_system=='darwin'",
```
cbrnr, line 30:
```suggestion
    "pyobjc-framework-cocoa >= 10.0; platform_system=='Darwin'",
```

cbrnr: The value of `platform_system` is case-sensitive and starts with an upper case letter, so this must not be changed.

src/mnelab/utils/channelstats_calculations.py
cbrnr, line 1: Can you move the function into `utils.py`? I don't think we need a separate module just for this functionality. Also, please add `calculate_channel_stats` to `__init__.py` so that it can be imported with `from mnelab.utils import calculate_channel_stats`.

src/mnelab/utils/channelstats_calculations.py
```diff
@@ -0,0 +1,39 @@
...
+        ch = raw.info["chs"][i]
+        data = raw[i][0]
+        cols["type"].append(channel_type(raw.info, i))
+        cols["unit"].append(ch.get("unit", ""))
+        cols["min"].append(np.min(data))
+        cols["Q1"].append(np.percentile(data, 25))
+        cols["mean"].append(np.mean(data))
+        cols["median"].append(np.median(data))
+        cols["Q3"].append(np.percentile(data, 75))
+        cols["max"].append(np.max(data))
```
cbrnr, line 27: It takes quite some time until the dialog window pops up, presumably because the calculation here uses a loop over channels. Can you use vectorized NumPy calls instead, e.g., if `data` is the entire 2D NumPy data array, you can do `np.mean(data, axis=1)` to get the mean for all channels simultaneously.

src/mnelab/dialogs/channel_stats.py
```diff
@@ -0,0 +1,95 @@
...
+            return str(self.value) < str(other.value)
+
+
+class ChannelStats(QDialog):
+    def __init__(self, parent, raw):
+        super().__init__(parent=parent)
+
+        # window
+        self.setWindowTitle("Channel Stats")
+        self.resize(650, 550)
```
cbrnr, line 34: It would be nice if the default size of the dialog was wide enough to avoid showing the horizontal scrollbar.

cbrnr:
Running `np.percentile()` is still not particularly fast, but I don't think that this can easily be sped up further, so we'll have to live with some delay.

cbrnr:
Thanks @bkloeckl!

cbrnr:
@larsoner you are a NumPy/SciPy expert, do you know why the `np.percentile()` function is relatively slow (~1.5s on medium-sized data, i.e., ~70 channels times ~1.3 million samples)? Is this normal? The code is here: https://github.com/bkloeckl/mnelab/blob/c7be862a673914fc04feb602b0cafd5348aaf846/src/mnelab/utils/utils.py#L39-L65

larsoner:
IIRC percentile does sort() under the hood, does that take about the same amount of time?

cbrnr:
For some reason, `sort()` seems to be even slower (unless I made a mistake). I'll try to play around with it a bit (and I'll create a MWE to rule out any Qt-related things).

cbrnr:
Yeah, sorting is more than twice as slow it seems:

```python
import time

import numpy as np

rng = np.random.default_rng(42)
data = rng.random(size=(70, 1_500_000))

start = time.time()
np.percentile(data, [0, 25, 50, 75, 100], axis=1)
print(time.time() - start)

start = time.time()
np.sort(data, axis=1)
print(time.time() - start)
```

Output on my machine:

```
1.5588321685791016
3.7198448181152344
```

Issue #385, opened 2023-08-12
read_raw_xdf(): resampling assumes evenly sampling


DominiqueMakowski:
I was adding [a wrapper](https://github.com/neuropsychology/NeuroKit/pull/874) around pyxdf to load and tidy up xdf data in Neurokit when I stumbled upon these issues:

https://github.com/xdf-modules/pyxdf/issues/79
https://github.com/xdf-modules/pyxdf/pull/1

So I decided to give mnelab a try to see if what I was doing was correct, but I found an issue that happens when there was interruptions in the streaming:

```python
import neurokit2 as nk

df, info = nk.read_xdf(
    "raw/physio/sub-01/ses-S001/eeg/sub-01_ses-S001_task-HCT_run-001_eeg.xdf"
)

from mnelab.io.xdf import read_raw_xdf

df2 = read_raw_xdf(
    "raw/physio/sub-01/ses-S001/eeg/sub-01_ses-S001_task-HCT_run-001_eeg.xdf",
    stream_ids=[1, 2, 3, 4, 5],
    fs_new=2000,
).to_data_frame()
df2.index = df.index

plt.close()
fig, ax = plt.subplots(nrows=3)
for c in ["AF7"]:
    ax[0].plot(info["data"][3][c], label="original", color="blue")
    ax[1].plot(df[c], label="NK", color="green")
    ax[2].plot(df2[c], label="mnelab", color="orange")
plt.legend()
plt.show()
```
![image](https://github.com/cbrnr/mnelab/assets/8875533/f615a1d7-d6b7-497a-ba02-6323dec5b548)

Essentially I plotted:

- in blue, the original signal from Muse EEG sampled at ~250He
- in orange, the resampling at 2000 Hz by NK based on pandas that is timestamp-based
- in blue the one by mnelab

There is an interruption at the beginning at the stream, but mnelab probably interpolates linearly which distorts the whole thing. 

I can send you that xdf file by email if you need :)

(also tagging my friend @pmavros as this might be relevant to our eeg processing)

cbrnr:
Thanks @DominiqueMakowski, you are correct that we currently assume regularly spaced samples per stream. Using pandas to handle interpolation is actually very clever, I wonder why I haven't thought of it before 😆 (the tradeoff is of course that it is a rather large dependency, but maybe worth it even for MNELAB).

Could you share the file with me so that I can play around with it to get a better grasp of the problem?

[...]

PR #467, opened 2024-12-26
Load .xdf files with gaps correctly

bkloeckl:
Fix [#385](https://github.com/cbrnr/mnelab/issues/385). Correctly load xdf files with incomplete data/gaps by inserting nan for missing data instead of interpolation.

### DIFF
--- a/src/mnelab/io/xdf.py
+++ b/src/mnelab/io/xdf.py
@@ -14,7 +14,7 @@ class RawXDF(BaseRaw):
     """Raw data from .xdf file."""
 
     def __init__(
-        self, fname, stream_ids, marker_ids=None, prefix_markers=False, fs_new=None
+        self, fname, stream_ids, marker_ids=None, prefix_markers=False, fs_new=None, gap_threshold=1.0
     ):
         """Read raw data from .xdf file.
 
@@ -41,6 +41,9 @@ class RawXDF(BaseRaw):
 
         streams, _ = load_xdf(fname)
         streams = {stream["info"]["stream_id"]: stream for stream in streams}
+
+        # Use these functions after loading the data
+        
 
         if all(_is_markerstream(streams[stream_id]) for stream_id in stream_ids):
             raise RuntimeError(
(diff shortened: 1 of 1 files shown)


src/mnelab/io/xdf.py
```diff
@@ -15,7 +15,13 @@ class RawXDF(BaseRaw):
...
 
     def __init__(
-        self, fname, stream_ids, marker_ids=None, prefix_markers=False, fs_new=None
+        self,
+        fname,
+        stream_ids,
+        marker_ids=None,
+        prefix_markers=False,
+        fs_new=None,
+        gap_threshold=1.0,
```
cbrnr, line 24: Can you add a docstring entry please?

cbrnr:
There seems to be something wrong when resampling the data, but the problem also exists on `main`. When loading the file [`sub-13_ses-S001_task-HCT_run-001_eeg.xdf`](https://osf.io/6y8ub), the signal scaling (at least for the EEG channels) changes as well as the location of the gap:

Using the default of 52 Hz:
52Hz

Using 256 Hz:
256Hz

It looks like the first one shows the correct signals. I'm not sure what the problem is, because the streams are always resampled (this is necessary when loading 2 or more streams).

DominiqueMakowski:
I've tested with [NK](https://github.com/neuropsychology/NeuroKit/blob/master/neurokit2/data/read_xdf.py) and I it works "as expected"

```python
import neurokit2 as nk

df, sr = nk.read_xdf("sub-13_ses-S001_task-HCT_run-001_eeg.xdf", upsample=0.25)
df[["TP9", "TP10", "LUX", "ECGBIT1"]].iloc[0:60000].plot(subplots=True)
```
image

```python
df, sr = nk.read_xdf("sub-13_ses-S001_task-HCT_run-001_eeg.xdf", upsample=0.5)
df[["TP9", "TP10", "LUX", "ECGBIT1"]].iloc[0:120000].plot(subplots=True)
```



The scaling issue is quite odd, but probably related to some internal handling

cbrnr:
Wait, but here you're using your own (=NK) implementation, right? You did not compare it to what MNELAB is doing in this PR?

DominiqueMakowski:
well no but to be honest I don't really have a good grasp on what mnelab does for xdf 😕

Issue #315, opened 2022-02-20
Add option to read custom montage files


cbrnr:
I think it would be great if users could read custom montage files (in addition to the built-in ones). This could be integrated into the existing dialog.

PR #468, opened 2024-12-26
Custom montage

bkloeckl:
Fix [#315](https://github.com/cbrnr/mnelab/issues/315). Allow users to select custom montage file in montage dialog


cbrnr:
Thanks @bkloeckl!

Issue #259, opened 2022-02-07
Extend filter dialog


cbrnr:
The filter dialog could use more options. For example, a notch filter is very often required, but currently only a highpass, lowpass, or bandpass filter can be applied.

cbrnr:
Implemented in #469.

PR #469, opened 2024-12-27
Extend filter dialog

bkloeckl:
Fix[ #259](https://github.com/cbrnr/mnelab/issues/259). Extend filter dialog by notch filter and savgol filter options.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ### Added
 - Add close icon to sidebar on hover ([#454](https://github.com/cbrnr/mnelab/pull/454) by [Benedikt Klöckl](https://github.com/bkloeckl))
 - Add Python 3.13 support ([#457](https://github.com/cbrnr/mnelab/pull/457) by [Clemens Brunner](https://github.com/cbrnr))
+- Add new filter dialog options for notch and savgol filters ([#469](https://github.com/cbrnr/mnelab/pull/469) by [Clemens Brunner](https://github.com/cbrnr))
 
 ### Changed
 - Change the append dialog appearance to include original indices used for identifying the data ([#449](https://github.com/cbrnr/mnelab/pull/449) by [Benedikt Klöckl](https://github.com/bkloeckl))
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - Add Python 3.13 support ([#457](https://github.com/cbrnr/mnelab/pull/457) by [Clemens Brunner](https://github.com/cbrnr))
 - XDF reader now parses measurement date ([#470](https://github.com/cbrnr/mnelab/pull/470) by [Stefan Appelhoff](https://stefanappelhoff.com))
 - Add support for loading custom montage files ([#468](https://github.com/cbrnr/mnelab/pull/468) by [Benedikt Klöckl](https://github.com/bkloeckl) and [Clemens Brunner](https://github.com/cbrnr))
+- Add new filter dialog options for notch and savgol filters ([#469](https://github.com/cbrnr/mnelab/pull/469) by [Benedikt Klöckl](https://github.com/bkloeckl))
```
cbrnr, line 7: Since there is no savgol option anymore, please adapt the changelog entry.

src/mnelab/dialogs/filter.py
```diff
@@ -3,34 +3,107 @@
...
+        self.bandpass_button.setChecked(True)
+        self.lowpass_button = QRadioButton("Lowpass")
+        self.highpass_button = QRadioButton("Highpass")
+        self.notch_button = QRadioButton("Notch")
+
+        self.filter_group = QButtonGroup(self)
+        self.filter_group.addButton(self.bandpass_button)
+        self.filter_group.addButton(self.lowpass_button)
+        self.filter_group.addButton(self.highpass_button)
+        self.filter_group.addButton(self.notch_button)
```
cbrnr, line 35: Can you put these two `QButtonGroup`s into two `QGroupBox` containers? That way, the dialog would be a bit more readable, because you can add a title for each `QGroupBox` (e.g., "Filter type" and "Filter Settings").

src/mnelab/dialogs/filter.py
cbrnr, line 1: Currently, the dialog can be accepted (by pressing the OK button) even if no frequencies are entered. Please add some sanity checks so that mandatory filter parameters are present (and otherwise disable the OK button). For the bandpass filter, you should also check if the lower frequency is really less than the upper frequency.

Also, the input fields currently accept any character, but you could limit it to numeric values by using a [QDoubleSpinBox](https://doc.qt.io/qt-6/qdoublespinbox.html).

cbrnr:
Thanks @bkloeckl!

PR #470, opened 2025-01-07
ENH: XDF reader: parse meas_date

sappelhoff:
This will allow the XDF reader to parse the measurement date from the file.


sappelhoff:
> And thanks for the links, maybe the specs wiki should be updated then?

I agree! @cboulay what do you think? Do you want to update the Wiki, given that you wrote that PR in 2020?

src/mnelab/io/xdf.py
```diff
@@ -104,6 +105,18 @@ def __init__(
             ]
             self.annotations.append(onsets, [0] * len(onsets), descriptions)
 
+        recording_datetime = header.get("info", {}).get("datetime", [None])[0]
```
cbrnr, line 108: I guess the `info` tag is mandatory, so no need to use `.get()`.

src/mnelab/io/xdf.py
```diff
@@ -104,6 +105,18 @@ def __init__(
...
+        if recording_datetime:
```
cbrnr, line 109:
```suggestion
        if recording_datetime is not None:
```

cboulay:
The addition of that field in LabRecorder is intentionally not documented in the XDF specs because it is not an XDF requirement. Making it a requirement would bump the spec version and also create divergence from the _other_ standard [here](https://standards.globalspec.com/std/10193987/CTA-2060).

The XDF header is eXtensible ML. If it's been extended by some tool you are welcome to look if that extension is there and use it if it is. But don't assume it is!

Additionally, I strongly encourage you to document that this field should not be relied upon for synchronization purposes. Even another XDF file recorded on the same computer might have been created after an NTP update so there will be some unknown offset separating the two files' datetimes.

cbrnr:
> from the other standard here.

👀 I have never heard of that one. Is it related to/derived from XDF? Is it intended to be compatible? I agree with your arguments, but deviating from some non-related format should not prevent us from continuing to evolve XDF.

sappelhoff:
> I have never heard of that one. Is it related to/derived from XDF? Is it intended to be compatible? I

Indeed, and it seems to be behind a paywall too 😬 😱 

> deviating from some non-related format should not prevent us from continuing to evolve XDF

agreed.

cboulay:
I believe it is EXACTLY the same as XDF, though I've never read it either. Deviating from that standard will upset some people from SCCN and potentially create divergences with industry users of this formal XDF standard. I don't think the new date time field in the header merits the complications a version bump will cause, especially considering:

* the header is designed to be extensible so its addition does not violate the spec
* the datetime field has low practical benefit -- it's only useful if you already know quite a bit about the data and you have to remove it when anonymizing anyway!

cbrnr:
@cboulay no worries, we're just trying to read that field if it exists. The discussion on XDF vs. the "industry standard" (air quotes on purpose) is not really relevant here (although in principle of course this is an interesting topic, because I assume we, the community, will eventually want to evolve XDF, so at some point there might be XDF >1.0, which is fine because the "standard" is based on XDF 1.0).

cbrnr:
Hmmm, I think we're running into a uv issue here (this is happening in all current PRs)...

cbrnr:
@sappelhoff do we know the exact format of the datetime string? If it is always provided as UTC, we could just remove the `z` character and parse it with `datetime.fromisoformat()` even with Python < 3.11. I'm asking because you are adding `python-dateutil` as an additional dependency.

sappelhoff:
The datetime string is formatted according to the ISO 8601 standard (%FT%T%z).

This is the code:

```
	std::time_t now = std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
	header << "\n    <datetime>" << std::put_time(std::localtime(&now), "%FT%T%z") << "</datetime>";
```

> I'm asking because you are adding python-dateutil as an additional dependency.

good point, I did not add it to the requirements. I am fine with implementing a workaround, but would adding this dependency also be fine until support for Python <3.11 is dropped?

cbrnr:
But wait, if that's the code that creates those datetime strings, meaning that the time zone gets appended as `+HH:MM` or `-HH:MM`, then unless I'm totally wrong, even Python versions < 3.11 should be able to parse them using `datetime.fromisoformat()` (at least that's what they claim in their [docs](https://docs.python.org/3.10/library/datetime.html#datetime.datetime.fromisoformat))?

sappelhoff:
👇 the example below contains a datetime string copy pasted from a recently recorded XDF file 🤔 

fails on Py < 3.11

image

succeeds on Py >= 3.11

image

apparently `datetime.fromisoformat()` prior to Python 3.11 has a problem with the `T` ... only this format is supported: `YYYY-MM-DD[*HH[:MM[:SS[.fff[fff]]]][+HH:MM[:SS[.ffffff]]]]`

So another problem is parsing the `0200` from the `+0200` suffix is parsed 🤔 it doesn't fit with `[+HH:MM[:SS[.ffffff]]]]` because it is in the format `[+HHMM]`

cbrnr:
That's really weird, because it works for me (on Python 3.10.6):

```
Python 3.10.16 (main, Jan  5 2025, 06:39:30) [Clang 19.1.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from datetime import datetime
>>> datetime.fromisoformat("2024-10-14T17:18:30+02:00")
datetime.datetime(2024, 10, 14, 17, 18, 30, tzinfo=datetime.timezone(datetime.timedelta(seconds=7200)))
```

sappelhoff:
🤷 This is of course up to you, but I think adding a temporary dependency on `dateutil` is cleaner 🤔

cbrnr:
> 🤷 This is of course up to you, but I think adding a temporary dependency on `dateutil` is cleaner 🤔

I'd probably go with the manual solution though, something like

```python
recording_datetime = recording_datetime[:-2] + ":" + recording_datetime[-2:]
```

Then everything works with all supported Python versions.

cbrnr:
Perfect, thanks @sappelhoff!

sappelhoff:
Thank you too for the discussion!

cboulay:
Not sure I groked this correctly. Did I make a mistake in the formatting of the datetime string in LabRecorder?

cbrnr:
Nah, I think that's fine, there are three options to specify the offset: https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators

PR #472, opened 2025-01-08
Exclude PySide 6.8.1.1

cbrnr:
This should fix CI errors with PySide 6.8.1.1 not being installable on non-macOS platforms.


cbrnr:
Missing wheels are now available on PyPI so this PR is not necessary anymore.

PR #479, opened 2025-02-18
Update mainwindow.py

sjentsch:
if self.recent is None (at the first start), MNELAB otherwise crashes with this error message:

[...]
  File "/opt/Software4EEG/MNE/lib/python3.12/site-packages/mnelab/mainwindow.py", line 78, in __init__
    self.recent = [recent for recent in self.recent if recent is not None]
[...]
TypeError: 'NoneType' object is not iterable

### DIFF
--- a/src/mnelab/mainwindow.py
+++ b/src/mnelab/mainwindow.py
@@ -74,7 +74,8 @@ class MainWindow(QMainWindow):
         self.installEventFilter(self)
 
         # remove None entries from self.recent
-        self.recent = [recent for recent in self.recent if recent is not None]
+        if self.recent != None:
+            self.recent = [recent for recent in self.recent if recent is not None]
 
         # plot backend
         self.plot_backends = ["Matplotlib"]


cbrnr:
Hmmm, I wonder why this happens, because if the setting is not available, it should be an empty list:

https://github.com/cbrnr/mnelab/blob/main/src/mnelab/settings.py#L21

I just tried deleting the settings and starting MNELAB, and it works as expected. Can you provide step by step instructions on how to reproduce this issue?

sjentsch:
Could you tell me where the settings are stored?
Unfortunately, I can't tell how the mix-up came about, I am not aware of having done something out of the ordinary, and it was the first time that I started MNELAB (at least with that user). May it be, that it happens when one opens MNELAB and then just closes it without having opened a file?

In any case, the if condition that I added doesn't hurt (I suppose), except from that it may be even better to add:
```
else:
    self.recent = []
```

cbrnr:
Settings are stored in a [platform-dependent location](https://doc.qt.io/qt-6/qsettings.html#locations-where-application-settings-are-stored); MNELAB uses the `NativeFormat` option, which means that since you seem to be on Linux, the settings should be stored in `$HOME/.config/cbrnr/MNELAB.conf` (if not, please refer to the `QSettings` documentation I linked).

Rather than just adding your proposed fix (which would solve the issue), I would like to first understand why this is happening. The problem might be in writing the settings rather than reading them.

I think we really need tests for the settings module, and these tests should be run on all three platforms (Windows, macOS, Linux).

sjentsch:
Hei Clemens,

you are likely right about that the error lies in writing. I attached you my config file, which , for some reason show `recent=@Invalid()`. I don't know what the supposed value for that is?

Thanks and cheers, Sebastian

PS: I had to change the extension, GitHub wouldn't accept .conf
[MNELAB.conf.txt](https://github.com/user-attachments/files/18884579/MNELAB.conf.txt)

cbrnr:
Yep, this should not happen! I'll look into it.

cbrnr:
I'm testing a better QSettings mechanism in #482, feel free to try it out, give feedback, and/or contribute over there (I'm closing this PR in favor of #482).

PR #482, opened 2025-02-20
Improve settings

cbrnr:
The settings location is now shown in the Settings dialog. Also, we need to test reading/writing settings on all three platforms (Windows, macOS, Linux) (this is still to do).


sjentsch:
Hei,

yes, it seems to work now. The config file now just contains:
```
[General]
pos=@Point(-11 -49)
size=@Size(700 500)
```
and there is no problem with opening it again afterwards.

Thanks for your help, Sebastian

PR #487, opened 2025-03-26
Change method of obtaining package versions

myd7349:
Hello!

While testing the ICA feature, I noticed that the About window indicated that `sklearn` was not installed. However, when I tried to install `sklearn` with:

```
uv add sklearn
```

uv displayed the following error message:.

```
      [stderr]
      The 'sklearn' PyPI package is deprecated, use 'scikit-learn'
      rather than 'sklearn' for pip commands.
```

This motivated this PR: `sklearn` is a deprecated package name and should not be used.

This PR uses `importlib.metadata.version`, introduced in Python 3.8, to obtain package versions instead of relying on `__version__`.  `version('scikit-learn')` also works well.  

The only "drawback" of this approach is that it depends on the information provided in the `.dist-info` directory, so these details need to be included when packaging with PyInstaller.

References:

- [How do I check the versions of Python modules?](https://stackoverflow.com/questions/20180543/how-do-i-check-the-versions-of-python-modules)
- https://docs.python.org/3/library/importlib.metadata.html#distribution-versions
- https://packaging.python.org/en/latest/specifications/recording-installed-packages/

### DIFF
--- a/src/mnelab/utils/dependencies.py
+++ b/src/mnelab/utils/dependencies.py
@@ -6,7 +6,7 @@ from importlib.metadata import version, PackageNotFoundError
 from itertools import chain
 
 required = ("mne", "PySide6", "edfio", "matplotlib", "numpy", "pyxdf", "scipy", "pybv")
-optional = ("mne-qt-browser", "picard", "scikit-learn")
+optional = ("mne-qt-browser", "python-picard", "scikit-learn")
 
 have = {}
 for package in chain(required, optional):
--- a/src/mnelab/utils/dependencies.py
+++ b/src/mnelab/utils/dependencies.py
@@ -2,20 +2,15 @@
 #
 # License: BSD (3-clause)
 
-from importlib import import_module
+from importlib.metadata import version, PackageNotFoundError
+from itertools import chain
 
-required = ["mne", "PySide6", "edfio", "matplotlib", "numpy", "pyxdf", "scipy", "pybv"]
-optional = ["mne-qt-browser", "picard", "sklearn"]
+required = ("mne", "PySide6", "edfio", "matplotlib", "numpy", "pyxdf", "scipy", "pybv")
+optional = ("mne-qt-browser", "picard", "scikit-learn")
 
 have = {}
-for package in required + optional:
+for package in chain(required, optional):
     try:
-        mod = import_module(package.replace("-", "_"))
-    except ModuleNotFoundError:
+        have[package] = version(package)
+    except PackageNotFoundError:
         have[package] = False
-    else:  # module successfully imported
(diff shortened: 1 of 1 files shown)


cbrnr:
Indeed you cannot install Scikit-Learn with `uv add sklearn` for two reasons: (1) the package is called `scikit-learn` and (2) `uv add scikit-learn` would just add it to `pyproject.toml` and not install it.

Do you start MNELAB with `uvx mnelabˋ? Then you could simply change that to `uvx "mnelab[full]"`.

myd7349:
Hello.

Today, when I tried to run mnelab via `uvx mnelab[full]` on a Windows 11 machine, I found that the installation failed. Both PowerShell and CMD reported syntax errors(I tried `uvx mnelab[full]` and `uvx "mnelab[full]"`).

However, when I tried the same commands on my personal computer after work, mnelab ran successfully.

An interesting thing is that after failing to run `uvx mnelab[full]` on the first computer, I tried installing the picard package separately. However, I discovered that `picard` is actually a different package (https://github.com/metabrainz/picard), while the one I needed was `python-picard` (https://github.com/pierreablin/picard, forgive me, I didn't check pyproject.toml directly at that time).

Therefore, for completeness, I will rename `picard` to `python-picard` as well in this PR. However, as you mentioned, this is not a major issue and does not affect the normal use of the software.

Finally, thank you for mnelab!

cbrnr:
Indeed, you just found another package with different package (`python-picard`) and import (`picard`) names. MNELAB is handling these cases correctly, so no need to make any changes.

Full installation should also work on Windows, can you post the exact error messages?

myd7349:
Hello. That computer is not with me right now, but I can take a screenshot tomorrow.  

In fact, on the first machine, running `uvx mnelab` caused three MessageBoxes to pop up sequentially, each reporting a *stdin/stdout/stderr redirect error*.  I guess this might be due to the Windows 11 version on that computer being somewhat outdated.

cbrnr:
These message boxes appeared with older versions of `uv`, so you might want to upgrade `uv` before trying again. Also, even though these message boxes popped up, MNELAB should start normally afterwards. But again, this problem is fixed with recent `uv` versions.

myd7349:
> These message boxes appeared with older versions of `uv`, so you might want to upgrade `uv` before trying again. Also, even though these message boxes popped up, MNELAB should start normally afterwards. But again, this problem is fixed with recent `uv` versions.

Thank you! I will give it a try tomorrow.

myd7349:
Hello. @cbrnr 

Just as you said, after upgrading uv (0.5.14 -> 0.6.10), problem solved. Thank you!

Issue #471, opened 2025-01-07
XDF reading: merging streams via resampling impacts channels that contain NaN


sappelhoff:
- Tobii eyetracker devices may produce time series data that contain NaN values (for example pupil dilation when the eye is closed)
- When streaming several Tobii eyetracker streams to LSL and recording in XDF, one may eventually want to merge the separate eyetracker streams into one MNE raw object
- In theory, this is possible with `mnelab.io.xdf.read_raw_xdf`
- **however**, the presence of NaN values in some channels of the streams result in the entire channel being NaN

I suspect this is due to the resampling done here:

https://github.com/cbrnr/mnelab/blob/4bbadaaa7e19d2a093c73c0c5afdc7110a5743c4/src/mnelab/io/xdf.py#L149

See this proof of principle:

```python
import numpy as np
import scipy.signal

x = np.arange(100) * 1.0
x[1] = np.nan
y = scipy.signal.resample(x, 90, axis=0)
assert np.isnan(y).all()  # this passes
```

I don't really have an idea how to prevent this. Would masking NaNs with "placeholder values" prior to resampling be an option? and then converting them back to NaN after resampling?

cbrnr:
Maybe some domain-specific preprocessing could be an option? Would using the last value instead of NaN make sense? I think some sort of interpolation will be necessary before resampling.

sappelhoff:
> Maybe some domain-specific preprocessing could be an option? 

Yes, a current solution is to read each stream separately into mne Raw objects (via mnelab), then do the processing, then resample and crop the raw objects, and finally call `raw.add_channels()` to concatenate the objects

Perhaps this is the only sensible way to avoid unnecessarily blowing up the reading code here with an edge case 🤔

sappelhoff:
> I mean, if this is a common operation for eyetracking streams it might be worth including it as an option.

Yes, unlike in EEG or other time series, NaN values are to be expected in eye tracking, because when the tracker loses the eyes (e.g., due to a blink), a NaN value gets assigned.

>  If it's really an edge case, not so much.

I thought the edge case here is probably having two separate eye-tracking streams that need to be merged. Most people will only have one eye-tracking stream, and will thus only have to resample if they want to combine it with another stream.

But then again, many users MIGHT want to combine their eye-tracking and EEG streams. So perhaps this is all just a valid case.

>  Can you provide more details on how you actually "do the processing" for these kinds of signals? How do you get rid of the NaNs?

[...]

PR #510, opened 2025-09-24
Feature interpolate missing

behinger:
I use this PR in my lab to fix issues with breaks in the transmission (e.g. amplifier dies). This PR is heavily based on @bkloeckl work #467 - sorry for making a new PR, my git skills are not great enough to base mine on yours, probably there is a way to have you in here as the main contributor? I did change quite a bit the logic and placing of things. It is now also not breaking but optional.

- Semi-breaking is the change to actually push the keyword arguments of read_raw_xdf to the RawXDF object, which was probably a bug in the first place not to do so.

- I'm not sold on the specific interpolator, but my reading suggested this one should be good

- I havent added tests yet, I'm not super familiar with python test cases :S, but this code is now laying around for 2-3 months and I need to share it in the first place. 

This relates to #471 #385

### DIFF
--- a/src/mnelab/io/xdf.py
+++ b/src/mnelab/io/xdf.py
@@ -15,7 +15,7 @@ class RawXDF(BaseRaw):
     """Raw data from .xdf file."""
 
     def __init__(
-        self, fname, stream_ids, marker_ids=None, prefix_markers=False, fs_new=None
+        self, fname, stream_ids, marker_ids=None, prefix_markers=False, fs_new=None,        gap_threshold=0.1,interpolate_or_resample="interpolate"
     ):
         """Read raw data from .xdf file.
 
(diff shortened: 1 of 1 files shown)


cbrnr:
Thanks! Can you fix the style errors? I can't promise when I will have time to review this though. This has been a very complex topic, which is also why I think the previous solution by @bkloeckl was not ideal. In any case, this needs careful review, which means I need to take my time to think about the proposed fix.

cbrnr:
Also, can you test this PR with the following data sets taken from #385 (they contain long gaps):

1. https://osf.io/uc7wn/files/p9jz3
2. https://osf.io/uc7wn/files/6y8ub

cbrnr:
Wait, I tested this some more, and resampling is, as you correctly say, not working with gaps. More precisely, it doesn't give an error, but it incorrectly shifts the data after the gap right to the end of the first segment. Even worse, if `gap_threshold > 0`, some of the actual data gets nan'd out to create the illusion of a gap. So I think we should probably always use interpolation (with or without gap detection). Make sense?

behinger:
I was about to write exactly that - your answer gave me the right hints. The issue is that `resample` assumes everything is nicely sampled, e.g. you don't put in the original timestamps.

---

Resamp vs. interpolation: I think you had a discussion at some point about interpoaltion & potential aliasing somewhere else with someone else on the LSL/xdf-universe. I haven't thought these steps too much - for me I argued in my head, that I'm only interpolation from e.g. 1000.01Hz to 1000Hz, so aliasing is not really a big issue. But if you want to downsample in the same step as interpolating, I don't know anymore.

That being said, maybe one could filter the data gap-wise, after gap detection, but prior to interpolation?

cbrnr:
I've now integrated this feature into the XDF dialog:

Screenshot 2025-10-22 at 12 31 51

By default, resampling is disabled. I think this is the best default, as it imports the original stream data as is. Note that gap detection is disabled, because it requires resampling (or rather, interpolation).

If you check the resampling checkbox, gap detection becomes available (but by default is unchecked). This means that the stream data will be resampled and any gaps will be ignored. If you check the gap detection checkbox (and choose a suitable threshold, the default is 0.1 s), the stream data will be interpolated and gaps larger than the threshold will be handled correctly.

You can test the current implementation in this PR with:

```
uvx --from git+https://github.com/behinger/mnelab@feature-interpolateMissing mnelab
```

PR #530, opened 2025-11-12
fix: Replace deprecated arguments in plotting functions

SchellanderF:
This PR addresses two deprecations in the visualization module:

1.  In `plot_erds`: Replaces the deprecated `vmin` and `vmax` arguments with the unified `vlim=(vmin, vmax)`.
2.  In `plot_evoked_topomaps`: Removes the deprecated `title` argument from the `plot_topomap` call and uses `fig.suptitle()` instead to set the title.


cbrnr:
Thanks @SchellanderF!

PR #531, opened 2025-11-14
Fix decimal separator in the drop epoch menu

SchellanderF:
This change updates the "Drop bad epochs" dialog to use QDoubleSpinBox, which correctly handles locale-aware decimal separators, aligning it with the other dialog windows. Additionally, the value-reading logic was also updated from float(widget.text()) to widget.value() to support this change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## [UNRELEASED] - YYYY-MM-DD
 ### 🔧 Fixed
-- Fixed a bug in the 'Drop bad epochs' dialog that did not accept commas as decimal separators ([531](https://github.com/cbrnr/mnelab/pull/531) by [Fabian Schellander](https://github.com/SchellanderF))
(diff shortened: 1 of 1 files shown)


SchellanderF:
> Apart from my inline comments, I have two general suggestions/concerns:
> 
> 1. I think we should not use the base SI units here, but the more practical ones like µV for EEG etc. (see https://github.com/cbrnr/mnelab/blob/main/src/mnelab/utils/utils.py#L54-L55 on how to get these units and scalings). Then we could limit the number of decimals and set the step size to something reasonable like 0.1.

I completly agree. I will get started on updating the dialog.

> 2. Previously, it was possible to have an empty text field, which resulted in the threshold not being used. This was probably not a good idea in the first place, but now the default value is zero, which is not a good default I think, because people should still get something half useful if they do not change the default. However, with a PTP of zero, all epochs will be rejected and none of them will be marked as flat. I am not sure how to best choose a default threshold for each signal type, but at least for EEG we could use 100µV for the maximum and 0.1µV for the minimum detection, but we should discuss this in more detail. Also, I think these options should be much better explained in the dialog, right now it is very difficult to figure out what these things do.

Would you suggest a dictionary which contains all the values (stepsizes, initial values etc) for the different types of signals? 
For the explanation, would adding a setToolTip() be sufficient or are you thinking of a more visible description label directly in the dialog?

PR #534, opened 2025-11-24
Add dialog to create annotations from event intervals

SchellanderF:
This PR introduces a new dialog window for creating annotations from events.

The dialog offers two modes:
1. **Copy Events to Annotations:** The existing functionality, moved into this dialog.
2. **Define Intervals between Events:** A new feature to create annotations based on start/end events.

### DIFF
--- a/src/mnelab/utils/utils.py
+++ b/src/mnelab/utils/utils.py
@@ -80,9 +80,9 @@ def annotations_between_events(
 ):
     """Create annotations between events.
 
-    This function identifies intervals starting with one of the `start_events`
-    and ending with the next occurrence of one of the `end_events`. Additionally, it
-    can also automatically create annotations to cover the beginning and end of the
+    This function identifies intervals starting with one of the `start_events` and
+    ending with the next occurrence of one of the `end_events`. Additionally, it can
+    also automatically create annotations to cover the beginning and end of the
     recording.
 
     Parameters
@@ -98,21 +98,21 @@ def annotations_between_events(
     annotation : str
         The description (label) to assign to the created annotations.
     max_time : float | None
-        The total duration of the data in seconds. Required if `extend_end`
-        is True to define the end of the recording. Defaults to None.
+        The total duration of the data in seconds. Required if `extend_end` is True to
+        define the end of the recording. Defaults to None.
     start_offset : float
         The offset in seconds to apply to the start events. Defaults to 0.
     end_offset : float
         The offset in seconds to apply to the end events. Defaults to 0.
     extend_start : bool
-        Whether to extend the first annotation to the start of the recording.
-        Defaults to True.
+        Whether to extend the first annotation to the start of the recording. Defaults
+        to True.
     extend_end : bool
-        Whether to extend the last annotation to the end of the recording.
-        Defaults to True.
+        Whether to extend the last annotation to the end of the recording. Defaults to
+        True.
     Returns
     -------
-    annotations : mne.Annotations
+    mne.Annotations
         The generated annotations object containing the annotated intervals.
     """
     onsets = []
--- a/src/mnelab/mainwindow.py
+++ b/src/mnelab/mainwindow.py
@@ -1209,7 +1209,7 @@ class MainWindow(QMainWindow):
                         f"    start_offset={interval_data['start_offset']},\n"
                         f"    end_offset={interval_data['end_offset']},\n"
                         f"    extend_start={interval_data['extend_start']},\n"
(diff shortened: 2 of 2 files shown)


cbrnr:
In addition, I was wondering if we could extend the functionality by allowing multiple start and end events. The reason for this is that often there are different event types depending on the response (e.g., 3 = correct, 4 = incorrect, 5 = missing), and users might still want to create annotations between any response (i.e., 3, 4, and 5 in this example) and a trial start (e.g., 1). WDYT?

CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 ## [UNRELEASED] - YYYY-MM-DD
+### ✨ Added
+- Add dialog to create annotations from events or event intervals ([#534](https://github.com/cbrnr/mnelab/pull/534) by [Fabian Schellander](https://github.com/SchellanderF))
```
cbrnr:
```suggestion
- Add dialog to create annotations from events or between specific events ([#534](https://github.com/cbrnr/mnelab/pull/534) by [Fabian Schellander](https://github.com/SchellanderF))
```

src/mnelab/dialogs/__init__.py
```diff
@@ -7,6 +7,7 @@
 from mnelab.dialogs.brainvision import BrainVisionDialog
 from mnelab.dialogs.calc import CalcDialog
 from mnelab.dialogs.channel_properties import ChannelPropertiesDialog
+from mnelab.dialogs.create_annotations_from_events import AnnotationsIntervalDialog
```
cbrnr: I think calling the module `annotations_from_events` is sufficient.

```suggestion
from mnelab.dialogs.annotations_from_events import AnnotationsIntervalDialog
```

SchellanderF:
I have updated it to match the other dialog windows. It is a standard QListWidget which supports multi-selection.

src/mnelab/dialogs/annotations_from_events.py
```diff
@@ -0,0 +1,210 @@
...
+
+        is_common = False
+        warning_msg = ""
+
+        # Check events
+        if has_start and has_end:
+            common_events = set(start_events).intersection(set(end_events))
+            if common_events:
+                is_common = True
+                warning_msg = "Error: Start and end events cannot be the same."
```
cbrnr:
```suggestion
                warning_msg = "Start and end events cannot be the same."
```

cbrnr (commented):
I like it! Regarding the history, it currently adds the following command:

```python
annotations_between_events(raw, events, {'start_event': [1], 'start_offset': 0.0, 'end_event': [3, 4, 5], 'end_offset': 0.0, 'annotation': 'BAD_INTERTRIAL', 'extend_start': True, 'extend_end': True})
```

A couple of suggestions/questions:

1. This function needs to be imported at the top of the history script.
2. `raw` is named `data` in the history.
3. Would it be possible for the function to have multiple regular parameters (with defaults where possible) instead of getting all that information via a dict?
4. Instead of passing `raw`, could the function return the `Annotations` object, which we could then set via `raw.set_annotations(...)`? (In the history, it would be `data.set_annotations(...)`.)

Also, the dialog now does not warn if the number of start/end events do not match. I think this is fine, but I just wanted to make sure that you did not accidentally drop that (it's still in the code).

src/mnelab/utils/utils.py
```diff
@@ -65,6 +66,91 @@ def calculate_channel_stats(raw):
...
+    end_events,
+    max_time=None,
+    start_offset=0,
+    end_offset=0,
+    annotation="BAD_INTERTRIAL",
+    extend_start=True,
+    extend_end=True,
+):
+    """Create annotations between events."""
+    onsets = []
```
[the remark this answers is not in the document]
cbrnr: Also, I think the `annotation` parameter should not have a default.

src/mnelab/dialogs/annotations_from_events.py
```diff
@@ -0,0 +1,210 @@
...
+        # checkboxes
+        self.to_start_check = QCheckBox("Extend to start")
+        self.to_start_check.setChecked(True)
+        self.grid.addWidget(self.to_start_check, 5, 0, 1, 2)
+
+        self.to_end_check = QCheckBox("Extend to end")
+        self.to_end_check.setChecked(True)
+        self.grid.addWidget(self.to_end_check, 6, 0, 1, 2)
+
+        # warning Label
```
cbrnr:
```suggestion
        # warning label
```

PR #543, opened 2025-12-18
Add ICLabel classification

SchellanderF:
This PR integrates ICLabel to classify ICA components

**Changes**
- Added a new dialog for displaying the class probabilities
- Implemented the calculation of the features (topoplot, autocorrelation and PSD) 
- Integrated the ONNX model to perform the classification

~~New Dependency: `onnxruntime`~~


cbrnr:
I also wonder if we should make the IClabel functionality an optional dependency...

cbrnr (commented):
I also made some superficial style-based comments. Edit: Or maybe this didn't work, I used the VS Code extension, but should probably have done it directly in the browser. Will do that again later, but feel free to start addressing the points I already mentioned.

SchellanderF:
Great ideas, I'll start implementing these fixes.

On the checkbox centering: I was following the style of the `Channel Properties` dialog. Do you want me to go back and center the checkboxes there as well?

src/mnelab/mainwindow.py
```diff
@@ -1119,6 +1132,28 @@ def apply_ica(self):
...
+                self.model.history.append(f"ica.exclude = {ica.exclude}")
+                self.data_changed()
+            except ValueError as e:
+                msgbox = ErrorMessageBox(
+                    self,
+                    "Could not label ICA components",
+                    str(e),
+                    traceback.format_exc(),
+                )
+                msgbox.show()
```
cbrnr: I wonder when this ValueError can occur – is this try/except really necessary?

cbrnr:
I like the tests you added, but instead of including the source code to generate the test data, I would put it in an actual Python script inside the test_data folder (you could name it `generate_test_data.py` or something like that). I would then mention in a comment at the top that `mne_iclabel` and `onnxruntime` are required to run that script (and therefore generate the data).

Also, I would rename `test_data` to just `data`, as this folder is already located inside `tests`.

PR #545, opened 2026-01-12
Add IC property visualization to ICLabel dialog

SchellanderF:
This PR improves the `ICLabelDialog` by integrating a detailed component property visualization. The implementation extends the standard `mne.ica.plot_properties` figure with a custom probability histogram.


cbrnr:
Thanks @SchellanderF! I've pushed a minor change which improves the look of the bar chart (hide x-axis ticks, draw the x-axis, move the text a tiny bit up, see screenshot below).

Also, I think we don't need a toolbar in this plot. I know that the original properties plot includes it, but I don't think it is necessary (and it takes up quite a lot of space). Would it be OK with you to remove it? In addition, can you also remove the large borders around the actual contents?

Screenshot 2026-01-13 at 09 03 43

cbrnr:
Can you push your changes?

PR #558, opened 2026-02-13
Add artifact detection dialog for epochs

SchellanderF:
Adds a new artifact detection dialog that allows users to detect and drop bad epochs using multiple detection methods.

- Support multiple detection methods: extreme values, peak-to-peak, kurtosis, AutoReject
- Interactive preview with manual epoch selection and visualization

New optional import: `autoreject`

TODO
-  Add detection info to history

### DIFF
--- a/src/mnelab/dialogs/artifact_detection.py
+++ b/src/mnelab/dialogs/artifact_detection.py
@@ -2,7 +2,6 @@
 #
 # License: BSD (3-clause)
 
-import mne
 from matplotlib.backends.backend_qtagg import FigureCanvas
 from matplotlib.colors import to_rgba_array
 from PySide6.QtCore import QEvent, Qt, QTimer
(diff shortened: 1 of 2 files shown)


cbrnr:
Thank you! I have played with the new dialog a bit and wanted to share some initial feedback (I haven't reviewed the code yet):

1. The QDoubleSpinBoxes should display only one decimal place, and their values should be right-aligned.
2. When no method is selected, I don't think we need to display a message. It should be clear that at least one method must be selected.
3. The info message should be left-aligned. If it is a full sentence, it should end with a period.
4. Currently, changing a parameter for one method triggers recomputation of all selected methods. It would be more efficient to recompute only the method whose parameter was modified.
5. When the backend is set to "Matplotlib", keyboard navigation in the browser (after clicking "View Epochs") does not work (e.g., arrow keys). This issue does not occur with the "Qt" backend.
6. I'm still struggling with the preview table a bit: a red ✗ indicates that a method marks an epoch as an artifact, while a green ✓ indicates it is clean. However, the "reject" column uses a check mark to indicate final rejection (i.e., that the epoch contains artifacts). Would it be an option to change the red ✗ to a red ✓ and do not show any symbol and color for epochs that a method did not mark as an artifact?
7. Finally, I was wondering if the dialog layout could follow the implementation of the "Change reference" dialog, where each method could have its own checkable group? I am not sure if this would be better, but it would make the two headings unnecessary. What's your opinion?

cbrnr:
Let's leave it as is for now then. My thinking was that if we add a method with more than one parameter in the future, the group boxes might be better.

cbrnr:
> I would go with always removing the mean and not exposing it as an option since the method is designed to catch signals that deviate from baseline.

Yes, perfect, I also prefer this approach!

cbrnr:
Very nice! There is only one (minor) thing remaining: Open the "Artifact Detection..." dialog, select at least one method, click on "Preview...", and then click on View Epochs... (this is the important part!). Now we don't want to do anything, so just close the plot, close the preview, and click "Cancel". This means we didn't drop anything, and this works. However, when closing MNELAB, the following message is printed in the console:

```
Dropped 64 epochs: 7, 8, 14, 16, 20, 24, 26, 33, 36, 37, 40, 42, 46, 52, 60, 64, 69, 78, 80, 82, 89, 92, 94, 100, 106, 108, 115, 116, 122, 124, 126, 128, 130, 140, 146, 148, 154, 156, 178, 182, 184, 188, 189, 192, 194, 198, 202, 204, 206, 208, 220, 221, 226, 230, 234, 238, 244, 245, 246, 254, 258, 260, 266, 282
The following epochs were marked as bad and are dropped:
[7, 8, 14, 16, 20, 24, 26, 33, 36, 37, 40, 42, 46, 52, 60, 64, 69, 78, 80, 82, 89, 92, 94, 100, 106, 108, 115, 116, 122, 124, 126, 128, 130, 140, 146, 148, 154, 156, 178, 182, 184, 188, 189, 192, 194, 198, 202, 204, 206, 208, 220, 221, 226, 230, 234, 238, 244, 245, 246, 254, 258, 260, 266, 282]
Channels marked as bad:
none
```

So even though we did not drop anything when MNELAB was running, these epochs seem to be dropped when MNELAB quits. This is not really a big deal, but it should be fixed nevertheless.

PR #565, opened 2026-02-25
Fix toolbar button hover on macOS

### LINKED ISSUE #460: Highlight active toolbar buttons on hover on macOS
On macOS, Qt doesn't highlight [toolbar buttons](https://doc.qt.io/qt-6/qtoolbutton.html#details) on hover (they call it "autoraise"), see [here](https://doc.qt.io/qt-6/qtoolbutton.html#autoRaise-prop). This is quite unfortunate, because (1) native macOS application do highlight on hover, and (2) from a usability point of view, it is really difficult to discover if a button is active or not.

Although this is not an MNELAB issue, maybe we could implement a workaround as outlined [here](https://stackoverflow.com/a/11852107/1112283) if it is not too hacky (e.g., it needs to support both light and dark themes).

cbrnr:
Fixes #460.

https://github.com/user-attachments/assets/53011b2c-0e4b-4e99-9c09-54f7476ab07c

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## [UNRELEASED] · YYYY-MM-DD
+### 🔧 Fixed
+- Fix toolbar button hover effects on macOS ([#565](https://github.com/cbrnr/mnelab/pull/565) by [Clemens Brunner](https://github.com/cbrnr))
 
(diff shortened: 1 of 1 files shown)


cbrnr:
We have to check how this looks on Linux and Windows though...

PR #573, opened 2026-02-27
Improve sidebar discoverability and selection visibility

### LINKED ISSUE #566: Sidebar color scheme causes UX issues
I just loaded the very first file and then the window looked like this:

Image

And I got really confused when looking at the left side of window, in that "side panel" where later all opened files will appear.

- I didn't know what that "0" meant – it's not obvious that I'm looking at some sort of a column layout here, as there are no column headers
- I didn't understand that the file that was loaded is right now "selected" / "active", as the highlight color is just the same as the rest of the interface and differs fundamentally from my system's highlight color for interaction modes like these:
    Image

Proposed improvements:
- Turn the sidebar content into an actual table with headers; this could even allow for sorting on click (ascending, descending)
- Use the system highlight color for the active element; if not possible, at least use a color that's different from the app window background color for differentiation

hoechenberger:
image

- Show column headers ("#" and "Name") in the sidebar table so the
layout and index column are self-explanatory.
- Also apply a stylesheet
that forces the system highlight color on selected rows, which was previously rendered as an inactive gray due to `Qt.NoFocus`.
- We keep `Qt.NoFocus` as I believe you chose this intentionally? To avoid issues with keyboard shortcuts perhaps?

Closes #566

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,13 @@
 ## [UNRELEASED] · YYYY-MM-DD
 ### ✨ Added
-- The dataset pane now contains badges that help you easily recognize the data type of each entry ([#576](https://github.com/cbrnr/mnelab/pull/576) by [Richard Höchenberger](https://github.com/hoechenberger))
+- The dataset pane now contains badges that make it easy to recognize the data type of each entry ([#576](https://github.com/cbrnr/mnelab/pull/576) by [Richard Höchenberger](https://github.com/hoechenberger))
 - Add mask argument to Find Events dialog, which is useful for parsing events in BDF files ([#584](https://github.com/cbrnr/mnelab/pull/584) by [Clemens Brunner](https://github.com/cbrnr))
 - The info pane now displays a copy button when hovering over the file name ([#578](https://github.com/cbrnr/mnelab/pull/578) and [#581](https://github.com/cbrnr/mnelab/pull/581) by [Richard Höchenberger](https://github.com/hoechenberger))
 
 ### 🔧 Fixed
 - Fix toolbar button hover effects on macOS ([#565](https://github.com/cbrnr/mnelab/pull/565) by [Clemens Brunner](https://github.com/cbrnr))
-- The dataset sidebar now contains badges that help you easily recognize the data type of each entry ([#576](https://github.com/cbrnr/mnelab/pull/576) by [Richard Höchenberger](https://github.com/hoechenberger))
-- Navigation between datasets by using the arrow up and down keys in the sidebar ([#573](https://github.com/cbrnr/mnelab/pull/573) by [Richard Höchenberger](https://github.com/hoechenberger))
-- Improve dataset sidebar styling on macOS ([#573](https://github.com/cbrnr/mnelab/pull/573) by [Richard Höchenberger](https://github.com/hoechenberger))
+- Add option to navigate between datasets in the sidebar with the up and down arrow keys ([#573](https://github.com/cbrnr/mnelab/pull/573) by [Richard Höchenberger](https://github.com/hoechenberger))
+- Improve sidebar styling ([#573](https://github.com/cbrnr/mnelab/pull/573) by [Richard Höchenberger](https://github.com/hoechenberger))
 - Improve dark mode theme for the Artifact Detection preview dialog ([#585](https://github.com/cbrnr/mnelab/pull/585) by [Fabian Schellander](https://github.com/SchellanderF))
 
 ### 🌀 Changed


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🔧 Fixed
 - Fix toolbar button hover effects on macOS ([#565](https://github.com/cbrnr/mnelab/pull/565) by [Clemens Brunner](https://github.com/cbrnr))
+- Improve sidebar styling on macOS ([#573](https://github.com/cbrnr/mnelab/pull/573) by [Richard Höchenberger](https://github.com/hoechenberger))
```
hoechenberger:
```suggestion
- You can navigate between datasets by using the arrow up and down keys in the sidebar ([#573](https://github.com/cbrnr/mnelab/pull/573) by [Richard Höchenberger](https://github.com/hoechenberger))
- Improve sidebar styling on macOS ([#573](https://github.com/cbrnr/mnelab/pull/573) by [Richard Höchenberger](https://github.com/hoechenberger))
```

src/mnelab/widgets/sidebar.py
```diff
@@ -44,6 +44,10 @@ def paint(self, painter, option, index):
         painter.setBrush(QColor(bg_hex))
         painter.drawRoundedRect(badge_rect, badge_h / 2, badge_h / 2)
 
+        painter.setPen(QColor(0, 0, 0, 40))
+        painter.setBrush(Qt.NoBrush)
+        painter.drawRoundedRect(badge_rect, badge_h / 2, badge_h / 2)
```
cbrnr: What does this do?

src/mnelab/widgets/sidebar.py
```diff
@@ -180,6 +209,11 @@ def style_rows(self):
             self.item(i, 1).setFlags(self.item(i, 1).flags() | Qt.ItemIsEditable)
             if self.item(i, 2) is not None:
                 self.item(i, 2).setFlags(Qt.ItemIsEnabled | Qt.ItemIsSelectable)
+        # After a rebuild (e.g. row removed), the cursor may already be hovering
+        # over a row without a MouseMove firing — update the close button immediately.
+        pos = self.viewport().mapFromGlobal(QCursor.pos())
+        index = self.indexAt(pos)
+        self.showCloseButton(index.row() if index.isValid() else -1)
```
cbrnr: Now the close button always shows when navigating to a dataset via up/down arrow keys.

Also, very minor point, but I'd like the comments to start with a lower-case letter...

src/mnelab/mainwindow.py
```diff
@@ -507,12 +509,15 @@ def data_changed(self):
         # update sidebar
         if len(self.model.data) > 0:
             self.sidebar.show()
+            # Block signals during rebuild: setRowCount(0) would otherwise
+            # emit currentCellChanged with row=-1, corrupting model.index.
```
cbrnr:
```suggestion
            # block signals during rebuild: setRowCount(0) would otherwise
            # emit currentCellChanged with row=-1, corrupting model.index
```

CHANGELOG.md
```diff
@@ -4,6 +4,13 @@
 
 ### 🔧 Fixed
 - Fix toolbar button hover effects on macOS ([#565](https://github.com/cbrnr/mnelab/pull/565) by [Clemens Brunner](https://github.com/cbrnr))
+- You can navigate between datasets by using the arrow up and down keys in the sidebar ([#573](https://github.com/cbrnr/mnelab/pull/573) by [Richard Höchenberger](https://github.com/hoechenberger))
+- The dataset sidebar now contains badges that help you easily recognize the data type of each entry ([#576](https://github.com/cbrnr/mnelab/pull/576) by [Richard Höchenberger](https://github.com/hoechenberger))
+- You can now navigate between datasets by using the arrow up and down keys in the sidebar ([#573](https://github.com/cbrnr/mnelab/pull/573) by [Richard Höchenberger](https://github.com/hoechenberger))
```
cbrnr: Please avoid "you" in the changelog. Also, it seems like you have to rebase/merge main.

hoechenberger:
> Finally, there is a minor visual issue. When a row is selected, you can actually click on the name or number again to select the cell. This highlights the cell further and should be disabled (just set the right selection mode).

On which platform is that? I cannot reproduce on macOS. Could you try with the current branch, perhaps it's something that has already been fixed? Thanks!

cbrnr:
> On which platform is that? I cannot reproduce on macOS. Could you try with the current branch, perhaps it's something that has already been fixed? Thanks!

This is on Linux/KDE, see the screencast (this is with the latest branch):

[Screencast_20260302_065206.webm](https://github.com/user-attachments/assets/af022a0b-afe5-46f7-9c93-f68eb8d77445)

cbrnr:
I tested on Windows and it works nicely! Do you want to add anything @hoechenberger? Otherwise I would go ahead and merge.

cbrnr:
Thanks @hoechenberger!

PR #627, opened 2026-05-07
Memory limitation

D3njo:


cbrnr:
I think we need to think about memory management some more (it would probably be easiest to add a setting where only the currently active data set is available in memory, everything else gets written to disk). Also, since this PR is not based on `main` it is very hard to tease apart the memory-related changes.

PR #630, opened 2026-05-11
Make manual theme switching more robust

cbrnr:
This should now work more reliably, but I still need to test this on all three platforms.

- [ ] Linux (KDE)
- [x] macOS ([standalone](https://github.com/cbrnr/mnelab/actions/runs/25661039311/artifacts/6914632901))
- [x] Windows ([standalone](https://github.com/cbrnr/mnelab/actions/runs/25661039311/artifacts/6914627368))

Testing should include:

- "Auto" should follow system theme changes
- "Light" should always use the light theme independent of the system theme
- "Dark" should always use the dark theme independent of the system theme

It is OK if the titlebar keeps following the system theme, but the palette and icons must adapt. Also, sometimes there seems to be some hiccups where restarting MNELAB resolves any issues; this is also OK.

On Linux/KDE, manual theme switching only works if using the Fusion theme. It does not work when using the Breeze theme, which is the default KDE theme, and in which case the MNELAB theme always follows the system theme ("Auto"). There are three options:

1. **Do nothing** — accept it as a platform limitation. The app still follows the system theme correctly (via `PaletteChange`), which is what most KDE users expect anyway. The option to manually set the theme should then probably be removed from Settings on this platform.
2. **Switch to the Fusion style** — `QApplication.setStyle("Fusion")` respects `setColorScheme()`, but it looks nothing like Breeze.
3. **Apply a full `QPalette` manually** — instead of calling `setColorScheme()`, construct and apply a complete light or dark `QPalette` directly on the `QApplication`. This bypasses the KDE platform theme's color management entirely. It's the most robust approach but requires defining all palette colors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,7 @@
 
 ### 🔧 Fixed
 - Fix the sidebar background color on Windows and Linux ([#622](https://github.com/cbrnr/mnelab/pull/622) and [#623](https://github.com/cbrnr/mnelab/pull/623) by [Clemens Brunner](https://github.com/cbrnr))
-
-### 🗑️ Removed
-- Remove manual Light/Dark theme options as this did not work reliably; the app now always follows the system theme ([#623](https://github.com/cbrnr/mnelab/pull/623) by [Clemens Brunner](https://github.com/cbrnr))
+- Make manual theme selection more robust ([#630](https://github.com/cbrnr/mnelab/pull/630) by [Clemens Brunner](https://github.com/cbrnr))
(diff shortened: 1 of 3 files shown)


cbrnr:
It's probably still not worth the trouble, especially since this doesn't work on all platforms. If there is strong demand for manual theme switching (independent of the system theme), we can reconsider adding it for supported platforms only (i.e., macOS and Linux, option 1 from above), but for now I'm closing it because light/dark mode following the system theme is probably the most important use case and is already working.

PR #635, opened 2026-05-11
Fix hamburger menu style

cbrnr:
Also use Cascadia Mono on Windows for monospace fonts and better kbd style. Also fix hamburger menu style (background and highlight from system theme).

Looks good on:

- [x] Windows
- [x] Linux (KDE) (works with Fusion but not with Breeze style, where right arrows overlap the menu text, see screenshot below)
- [x] macOS

image


cbrnr:
As a workaround, I moved the Settings action to the hamburger menu root:

image

